PR #7, opened 2017-07-24
Add Interferer as optional for MVDR filter

cksajil:
Interferer is added as optional for 'rake_mvdr_filters()' function

### DIFF
--- a/pyroomacoustics/beamforming.py
+++ b/pyroomacoustics/beamforming.py
@@ -1000,8 +1000,12 @@ class Beamformer(MicrophoneArray):
         response beamformer.
         '''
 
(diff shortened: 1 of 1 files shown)


pyroomacoustics/beamforming.py
```diff
@@ -1000,16 +1000,23 @@ def rake_mvdr_filters(self, source, interferer, R_n, delay=0.03, epsilon=5e-3):
         response beamformer.
```
fakufaku: Here, we should change the function arguments so that `interferer` and `R_n` are optional arguments with default value `None`, following the same pattern as `rake_perceptual_filters` for example.

It would also be nice to complete the docstring with parameters description.

cksajil: OK. Thank you for detailing the guidelines. I will follow the same.

PR #10, opened 2017-10-24
Fixes the missing import for polar_distance in doa.doa.polar_plt_dirac

jay-pee:
This is crucial when you want to use azimuth_ref arguemnt of the doa.doa.polar_plt_dirac function.

### DIFF
--- a/pyroomacoustics/doa/doa.py
+++ b/pyroomacoustics/doa/doa.py
@@ -2,6 +2,8 @@
 # Date: Feb 15, 2016
 from __future__ import division, print_function, absolute_import
 
+from pyroomacoustics.doa.utils import polar_distance
+
 """Direction of Arrival (DoA) estimation."""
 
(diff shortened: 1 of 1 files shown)


fakufaku:
Hi Philip,
Thanks for catching that. I have been using relative imports throughout the package. But following your pull request I was reading about that and it seems absolute vs relative is not completely clear. My understanding however is that as python 3, relative import are acceptable in a package and make it easier to rename things without breaking paths.
If you agree, would you consider amending your commit to use the relative path ? If not, I'd be happy to hear your point of view.
Cheers

jay-pee:
Hi Robin,
I am a starter in python and this is actually my first pull request, so maybe I not fully get what you mean.
I tried to correct the commit. Hope it's right.

fakufaku:
It looks all good. Thanks for the contribution!

PR #14, opened 2017-11-25
Same API for all analysis/process/synthesis use cases of STFT class

ebezzam:
Different cases include:
1. real-time: we know the size of our input and memory allocation is important so user should specify the 'num_frames' parameter to the STFT constructor
    a) single frame STFT --> 'num_frames' = 1
    b) multiple frame STFT --> 'num_frames' = <F>
2. non-real time: maybe we don't know the number of frames in our input and memory allocation is not critical so we can simply create the STFT object without the 'num_frames' parameter and let the STFT object allocate memory "on the fly" when we call analysis/synthesis/process


fakufaku:
This is starting to look nice. Thanks for the work! Just a couple of things:
There are other use cases to consider
3) Real-time but processing multiple frames at a time. In this case, it is still important to retain `fft_in_state`. That is, the chunks are still coming from the same stream, and it is important to retain the state for the reconstruction step.
4) Not real-time, or alternatively, we are doing STFT of multiple unrelated signals. In that case, we want to discard the state between two calls to `analysis`. Here, the STFT object is just a convenience to do easily multiple STFT with the same parameters (frame size, window, etc).
Maybe we could add a keyword to the constructor like `realtime` or `streaming` for example ? Maybe there is a way to distinguish also between these using the same mechanism as you describe above ?

fakufaku:
Or maybe, these are really just the same two use cases, so that if `num_frames` is specified, then we assume 3) and save the state between two calls, and when `num_frames` is not specified, then we discard the state.
In the end, I think we should avoid making too many implicit decisions for the user, so that we can have the following two keyword arguments:
* `num_frames` as you describe above
* `stateless` (=True or False)

fakufaku:
About the being explicit, there is another small modification that we could add to this pull req. Currently, if you select half-overlap, the constructor will change standard behavior and default to a hanning window. I think we should remove this and ask the user to specify a window explicitely when a rectangular window is not desired. (disclosure: I had some problems with that recently 😮 )

fakufaku:
And sorry about it, but maybe one last thing is that I quite liked better the order of axis you had before:
```
[frames, frequencies, channels]
```
Currently it is `[frequencies, frames, channels]`, which I think makes a little less sense. I think it is possible to simply change this by reversing the order of the elements of `new_shape`.

So just to summarize the changes requested:
1) Add a keyword `keep_state`
2) Add the logic to keep the state when `keep_state == True`
3) Remove the special behavior of using Hann window when `hop == nfft // 2`
4) Let `STFT.analysis` return `self.X` (this is just a reference to the transformed data)
5) Change `new_shape` so that the final axis order for `self.X` is `[frames, frequencies, channels]`

Thank you for bearing with me!!

ebezzam:
Ok finally got around to all of the changes you suggested...

1. Add a keyword `keep_state` --> I've called it `streaming` instead.
2. Add the logic to keep the state when `streaming == True` --> there are separate internal functions for this (`_analysis_streaming` and `_analysis_non_streaming`) and the appropriate one is called depending on the value of `streaming`. When `streaming==True` we would like an integer multiple of `hop` samples and when `streaming==False` we would like an integer number of `hop` samples + `block_size` samples.
3. Remove the special behavior of using Hann window when `hop == nfft // 2`
4. Let `STFT.analysis` return `self.X` (this is just a reference to the transformed data)
5. Change `new_shape` so that the final axis order for `self.X` is `[frames, frequencies, channels]`

I have a few questions/points:
1. I've combined the tests for all relevant use cases I could think of in `pyroomacoustics/realtime/tests/test_stft.py`. Should this be split up (almost 600 lines) or is it fine like this? There's one test case which I wasn't sure about adding because I don't think there's a use case for it (but maybe you could think of one). It would be the case (`num_frames=1, streaming=False,` and `hop`<`block_size`) or in other words we want to process one frame at a time and "they" are overlapping but we don't want to stitch between continuous calls of `analysis` and `synthesis`. I have the feeling it doesn't make sense to have overlapping frames since there's only one frame. Even the test case of (`num_frames=1, streaming=False,` and `hop`==`block_size`) seems like "overkill" for the `stft` block because it's pretty much doing the `block_size`-DFT of separate chunks of data, but I included this test case anyway.
2. I did not use the numpy `_as_strided` function for the case `streaming==True` but rather looping over the frames and extracting the appropriate samples (`_analysis_streaming` on Line 371). Should I also use `_as_strided` as I did for `_analysis_non_streaming` (Line 429)? I just need to think a little more on how to store samples correctly between separate calls of `analysis` and `synthesis`.
3. For the non-fixed case, I _append_ zeros to the input signal (Lines 288, 311, 326) so that we get the necessary numbers of samples for the streaming case (integer multiple of `hop` samples) or the non-streaming case (integer multiple of `hop` samples + `block_size`) depending on the value of `streaming`. I believe you discard the extra samples in the other `pyroomacoustics` implementation. Is it better to append zeros or discard the extra samples?

Let me know if you have questions or more points for improvement 😛

fakufaku:
@ebezzam Fantastic work! Maybe before we merge, we should think about renaming `realtime` -> `stft` and get rid of the original `stft.py` module ?
What do you think ?
And should we keep a convenience function that does the STFT in one line ? Rather than 1. creating object, 2. calling `process` ?

ebezzam:
@fakufaku thanks! so renaming the `realtime` folder (that contains `dft.py`, `stft.py`, and the tests) to `stft`? I think that would be more appropriate. And before removing the original `stft.py` module I would replicate something like the `spectrogram` function.

Yeah it might be good to have a one-shot function for convenience. Although the two-steps of creating the object and analysis separately is probably necessary for the "real time" case like for the `easy-dsp` examples and such applications where the memory allocation can be done in the initialization.

So that convenience function would both create the STFT object and apply the `analysis`+`process` function? And maybe also another convenience function for applying the synthesis.

fakufaku:
@ebezzam I think there is nothing worth saving in `stft.py` at this point :) Matplotlib has a nice spectrogram function and freqvec is available in numpy as `numpy.fft.fftfreq`. Let's just check that the interface for STFT is compatible with the old API.

The one shot function is nice when only one STFT/iSTFT is needed. Of cource when repeated calls are done, the object interface is better. Which is why we made one in the first place :)

ebezzam:
@fakufaku what should we check so that interface for STFT is compatible with the old API? 🙈  

I suppose I should also go through all the examples to see where the old stft may have been used?

ebezzam:
@fakufaku so I renamed the `realtime` submodule to `stft`. Instead of completely removing the original `stft.py` module, I copied its functions to the `stft/stft.py`. I renamed the one-shot functions `stft` and `istft` to `analysis` and `synthesis` so it matches the methods of the `STFT` class (I think there were some weird conflicts when I left the name `stft` because the module it's within and the file is called `stft`). I kept the `spectroplot` and `overlap_add` functions because they were being used by other examples/tests. Also I believe I went through all the examples and tests to change the function calls to the new API. Let me know what you think!

fakufaku:
Hey @ebezzam , I finally went back to this pull request. I have merged with master.

I have also decided to not rename the `realtime` module for now. I'd like to document the deprecation first. I have added deprecation warnings in the code and the doc for the `pyroomacoustics/stft.py` submodule. Let's have one version with the deprecation, then we can remove it in the next version.

What do you think ?

ebezzam:
Funny, I actually went back to this pull request over the weekend as well! I agree with having the renaming and deprecation in a different version/commit. I'd also change some internal things in `realtime.STFT`. 

One thing I was wondering is the dimension ordering. Before you said you had liked `[frames, frequencies, channels]` instead of `[frequencies, frames, channels]`. But would `[channels, frequencies, frames]` be alright?  With this ordering I could avoid a transpose of a possibly quite large matrix. This comes up because I use the `realtime.DFT` class in order to perform the DFT of multiple frames at once (of a single channel), but this class only supports the 2D transform for multiple columns so for data such as `[time,channels]` or `[time,frames]`. Another option would be to add an option to the DFT class so that it can also support the 2D transform for multiple rows at once. What do you think? 

I'm thinking this option for the DFT class would be good to add and wrt STFT dimension ordering, I'm open to whatever you think is best!

fakufaku:
Oh cool :)

For the axis ordering, I think it makes most sense from a memory ordering point of view. That is if  the multichannel data is stored interleaved. In `realtime.DFT`, can't you add an `axis` type of optional argument ?

fakufaku:
Another thing I am thinking about is the synthesis window! There is a well established condition for what the synthesis window should be to obtain the best reconstruction in case of non-linear processing ([Griffin & Lim 1984](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.306.7858&rep=rep1&type=pdf)). I think we should include this in the STFT. There are two ways:
1. We add a function that computes the synthesis window given the analysis window
2. We give analysis window and set `synthesis_window=True` to the `STFT` constructor and let it compute the synthesis window.
What do you think ?

ebezzam:
Yes I was thinking of adding an `axis` option for `realtime.DFT`, I'll make another PR with this.

By interleaved, you mean `channels` should be the last dimension?

That would be great for supporting STFT with overlaps different from 1/2. I'll take a look at the paper.

fakufaku:
> By interleaved, you mean channels should be the last dimension?
If using the C array ordering, yes.

>That would be great for supporting STFT with overlaps different from 1/2. I'll take a look at the paper.
I'm trying to implement it now. I am having some problems with inverse STFT using synthesis window in the non-streaming case. In that case, the beginning doesn't match.

fakufaku:
@ebezzam finally got around doing this :)

PR #28, opened 2018-05-01
Util for downloading and uncompressing tar.gz files.

ebezzam:
I seemed to have made a mess with when trying to rebase the commits....

I was trying to bring this branch of the fork back to a point without all the committed changes (by reverting) and then adding the function we created for downloading and uncompresssing tar.gz files, which is needed for the Google Speech Command dataset. 

@fakufaku let me know if we should squash these commits together (and if so how). What I was doing was squashing commits that were not part of`upstream` but then I would be asked to pull from `origin` which would then create more commits! Sorry for the troubles and thanks for your help.


fakufaku:
Not sure what happened there... 🤔 
This gave me motivation to rewrite this in a single function to handle both gz, bz2, tar and without the pesky temporary file. This is in commit 3814cab1. For now I have only tested with the bz2 file from CMU_ARCTIC. You can test it for the google dataset.

ebezzam:
Thanks! That's a better solution :)

PR #34, opened 2018-06-12
Locata dataset

fakufaku:
I added support for the [LOCATA challenge](http://www.locata-challenge.org) dataset. This is a dataset of real recordings where the goal is to locate static and moving sources using a few different microphone arrays.

### DIFF
--- a/pyroomacoustics/datasets/locata.py
+++ b/pyroomacoustics/datasets/locata.py
@@ -1,0 +1,228 @@
+'''
+The LOCATA Dataset
+==================
+
+This dataset was released as a challenge in spring 2018. The goal of the
+dataset is to provide a standard baseline for acoustic localization and
+tracking algorithm with both dynamic and static sources.
+
+The data is released under a very permissive license, however, at the time of
+writing of this code, the `official download <http://www.locata-challenge.org>`_
+page requires registration to obtain the dataset. No open mirror currently exists.
+
+License: `Open Data Commons Attribution License <https://opendatacommons.org/licenses/by/1-0/>`_
+
+Price: Free
+
+URL: http://www.locata-challenge.org
+'''
+import os, re
+import numpy as np
+from scipy.io import wavfile
+import datetime as dt
+import warnings
+
+from .base import Meta, AudioSample, Dataset
+
+url = 'http://www.locata-challenge.org'
+
+# the regexp for the dataset metadata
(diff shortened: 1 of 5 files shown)


pyroomacoustics/datasets/locata.py
```diff
@@ -0,0 +1,228 @@
...
+    This class wraps the LOCATA challenge dataset.
+
+    The development and evaluation datasets are released in folders with similar
+    names for task1 to task6.  We assume that the user will have saved these in
+    separate subfolder `eval` and `dev`.
+
+    Attributes
+    ----------
+    basedir: str, option
+        The directory where the CMU ARCTIC corpus is located/downloaded. By
```
ebezzam: change "CMU ARTIC corpus"

pyroomacoustics/datasets/locata.py
```diff
@@ -0,0 +1,228 @@
...
+        default, this is the current directory.
+    recordings: list of CMUArcticSentence
```
ebezzam: Same here

pyroomacoustics/datasets/locata.py
```diff
@@ -0,0 +1,228 @@
...
+        The list of all utterances in the corpus
+
+
+    Parameters
+    ----------
+    basedir: str, optional
+        The directory where the CMU ARCTIC corpus is located/downloaded. By
```
ebezzam: CMU

pyroomacoustics/datasets/locata.py
```diff
@@ -0,0 +1,228 @@
...
+        default, this is the current directory.
+    task: int, optional
+        The number of the task to read
+    rec: int or list of ints, optional
+        The recordings to consider
+    array: str or list of str, optional
+        The arrays to read in
+    dev: bool
+        Set to True to restrict to dev data, False to eval data. Not setting
```
ebezzam: Can write True/False as ``True``/``False`` so that it is formatted

pyroomacoustics/datasets/locata.py
```diff
@@ -0,0 +1,228 @@
...
+    path: str
+        The path to the recording folder
+    task: int
+        The task number
+    rec: int
+        The recording number
+    array: str
+        The array used
+    dev: bool
+        Wether this a development or evaluation recording
```
ebezzam: Whether

fakufaku:
@ebezzam Thanks for the comments. I tried to correct all that and improve some. Also checked the doc builds correctly.

PR #40, opened 2018-07-03
Renaming 'realtime' to 'transform'.

ebezzam:


fakufaku:
Hey Eric,

Thanks for opening this. I think for now we can keep both `transform` and `realtime` nomenclature. For that we can just create an alias in `pyroomacoustics.__init__` (e.g., `import .transforms as realtime`), and mark `realtime` as deprecated.

docs/pyroomacoustics.rst
```diff
@@ -20,12 +20,14 @@ Submodules
...
    pyroomacoustics.acoustics
    pyroomacoustics.beamforming
+   pyroomacoustics.build_rir
    pyroomacoustics.geometry
    pyroomacoustics.metrics
    pyroomacoustics.multirate
    pyroomacoustics.parameters
    pyroomacoustics.recognition
    pyroomacoustics.room
+   pyroomacoustics.scnr
```
fakufaku: Is this a new submodule ?

pyroomacoustics/__init__.py
```diff
@@ -126,7 +126,14 @@
...
 from . import doa
 from . import adaptive
-from . import realtime
+from . import transform
+from . import transform as realtime  # to be deprecated
 from . import experimental
 from . import datasets
 from . import bss
+
+import warnings
```
fakufaku: I know I asked for the deprecation warning. But is this the way to go ? The warning will be triggered everytime pyroomacoustics is loaded right ?

pyroomacoustics/transform/stft.py
```diff
@@ -31,7 +31,7 @@ class STFT(object):
     transform (optional) : str
         which FFT package to use: 'numpy' (default), 'pyfftw', or 'mkl'
     streaming (optional) : bool
-        whether (True) or not (False, default) to "stitch" samples between 
+        whether (True, default) or not (False) to "stitch" samples between
```
fakufaku: So what does making `streaming=True` the default mean practically ? Can we still just create the object and dump a large signal to change all at once ? I guess the answer is yes but, please confirm.

pyroomacoustics/transform/stft.py
```diff
@@ -31,7 +31,7 @@ class STFT(object):
     transform (optional) : str
```
fakufaku: I noticed that the doc doesn't render well. The argument name for optional arguments is missing. See [here](https://pyroomacoustics.readthedocs.io/en/latest/pyroomacoustics.realtime.stft.html).
I think the syntax should be

    argument_name: type, optional
        The description of the argument

fakufaku:
Well, to be fair, the problem with the doc was there before. So we can fix it here, or in another PR. As you see fit.

docs/pyroomacoustics.rst
ebezzam: oops I didn't realize that making the docs would create this. and yes I'm about to make a pull request with this soon.

ebezzam: should I remove it from this PR?

pyroomacoustics/__init__.py
ebezzam: According to here (https://docs.python.org/2/library/warnings.html). The warning will be ignored by default, as I added the "DeprecationWarning" label. Also it doesn't print out for me every time.

ebezzam: this is a better link to the table: https://docs.python.org/2/library/warnings.html#warning-categories

pyroomacoustics/transform/stft.py
ebezzam: Practically it means that consecutive calls of ``analysis``/``synthesis`` will by default stitch the consecutive frames (even if a particular call of ``analysis`` consists of multiple frames). ``streaming=False`` implies that consecutive calls of ``analysis``/``synthesis`` will not stitch the chunks together. I just noticed a problem with the ``stft_block_processing.py`` example where ``streaming`` was not set to ``True`` but it should have been. In any case, it's up for discussion whether we want this stitching to be done by default or not. I could also manually set ``streaming=True`` for the constructor in the ``stft_block_processing.py`` example.

pyroomacoustics/transform/stft.py
ebezzam: I'll change in this PR

docs/pyroomacoustics.rst
fakufaku: I guess it is better to include it with the relevant code.

pyroomacoustics/__init__.py
fakufaku: Cool. That sounds like a good way to do this.

PR #45, opened 2018-07-28
Add whitening method and unit test

jazcarretao:


pyroomacoustics/whitening.py
```diff
@@ -0,0 +1,40 @@
...
+    Returns
+    ----------
+    Y: ndarray (nframes, nfrequencies, nchannels)
+        STFT representation of the whitened signal
+    -------
+    '''
+
+    [n_frames, n_freq, n_chan] = X.shape
+    Y = np.zeros([n_frames, n_freq, n_chan], dtype=np.complex)
+    fudge = 1E-16
```
ebezzam: haha didn't know this term `fudge`! Maybe `tol` is more appropriate though. I think we've use that terminology in other part of the package.

pyroomacoustics/whitening.py
```diff
@@ -0,0 +1,40 @@
...
+    for f in range(n_freq):
+        Xi = X[:,f,:].T
+        cov_x = np.dot(Xi, np.conj(Xi).T)/n_frames
+        [eigenvalues, eigenvectors] = np.linalg.eigh(cov_x)
+        Y[:,f,:] = (np.dot(np.conj(eigenvectors).T, Xi).T/np.sqrt(eigenvalues) + fudge)
```
ebezzam: did you mean to add `fudge` to the denominator? to avoid division by numbers close to zero. Or to the final result (as is done now).

pyroomacoustics/tests/test_whitening.py
```diff
@@ -0,0 +1,36 @@
...
+    def test_whitening(self):
+        # Create multivariate distribution
+        mean = [0, 0]               # zero mean
+        covx = [[100, 5], [5, 2]]   # diagonal covariance positive-semidefinite, e.g., covx = [[100, 5], [5, 2]]
+        samples = 5000
+
+        # Test the input
+        assert np.all(np.linalg.eigvalsh(covx) >= 0), "Covariance matrix is not positive-semidefinite"
+        dimensions = len(covx)      # should be equal to two
+        assert dimensions == 2, "The number of dimensions must be equal to two"
```
ebezzam: I don't think this `assert` and the one above are necessary. It's not really testing the whitening and the conditions should be satisfied as you're manually setting the values.

pyroomacoustics/tests/test_whitening.py
```diff
@@ -0,0 +1,36 @@
...
+        X[:,:,1] = x1[:,None]
+
+        # Apply whitening
+        Y = pra.whitening(X)
+
+        # Test the output
+        covy = np.dot(Y[:,0,:].T,np.conj(Y[:,0,:])).T/samples
+        # Verify that the new correlation matrix is orthonormal
+        test = ortho_group.rvs(dimensions)
+        assert np.all(abs(np.dot(test,np.conj(test).T) - covy) < 1E-10), "Whitening unsuccessful"
```
ebezzam: I think this test could be made a bit more intuitive. We expect `covy` to be close to an identity matrix right? We could write something like this:
```
self.assertTrue(np.allclose(np.all(abs(np.eye(dimensions), covy))
```
This is also consistent with tests in other files, see [here](https://github.com/LCAV/pyroomacoustics/blob/master/pyroomacoustics/tests/test_stft_routines.py#L30).

You can also set the tolerance for `np.allclose`, see [here](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.allclose.html).

pyroomacoustics/tests/test_whitening.py
```diff
@@ -0,0 +1,36 @@
...
+
+if __name__ == '__main__':
```
ebezzam: The main function here is not necessary. You could do something like [here](https://github.com/LCAV/pyroomacoustics/blob/master/pyroomacoustics/transform/tests/test_dft.py#L112) where you create a utility function that performs the whitening for different parameters (e.g. dimension) so that in the main function you call this utility function to see that the unit test is satisfied when you run:
```
python test_whitening.py
```
And your `test_whitening` function should also call this utility function.

fakufaku:
@ebezzam Thanks for the thorough review and the introduction to test writing!
Concerning where to place the function, I think either `pyroomacoustics.utilities` or `pyroomacoustics.bss.common`. This will depend on whether the `whitening` function is useful for the several subpackages (e.g., `bss`, `doa`, etc), or only `bss`.
I think whitening appears in several areas, so maybe the former is more suitable.

Issue #18, opened 2018-02-08
Reverberation time


maelp:
Is there a simple way to convert between a reverberation time in T60 and an absorption coefficient?

This paper mentions the Eyring's empirical equation to convert to a reflection coefficient which I assume is (1 - absorption), is it something which is often used?

https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46107.pdf

fakufaku:
Hi @maelp , I don't know if you still need it, but you might give a go to [Sabine's formula](https://en.wikipedia.org/wiki/Reverberation#Sabine_equation). I was thinking myself that it would actually be nice to give an RT60 parameter rather than straight absorption coefficient when creating the room. Let's keep this issue open for now.

fakufaku:
Somewhat addressed by the addition of [inverse Sabine function](https://pyroomacoustics.readthedocs.io/en/pypi-release/pyroomacoustics.acoustics.html#pyroomacoustics.acoustics.inverse_sabine) in the latest version 0.4.0.

PR #46, opened 2018-07-30
Room from rt60

fakufaku:
Adds an option to `pyroomacoustics.room.ShoeBox` so that the reverberation time can be specified. The absorption and max_order are infered automatically using Sabine's formula.

This PR addresses issue #18

### DIFF
--- a/examples/room_rt60.py
+++ b/examples/room_rt60.py
@@ -1,0 +1,80 @@
+'''
+Room RT60
+---------
+
+In this example, we create a room with a pre-set reverberation time
+(according to Sabine's formula), and then check how the simulated RIR
+verifies the prediction.
+'''
+import math, itertools
+import numpy as np
+from scipy.signal import fftconvolve
+import pyroomacoustics as pra
+import matplotlib.pyplot as plt
+
+def rt60_analysis(room, mic, src, plot=True, rt60_tgt=None):
+    '''
+    Plot and analyze the RT60 of one of the RIR in the room.
+    '''
+
+    # Remove coefficients that might cause problems with the log
+    h = room.rir[mic][src]
+    T = np.arange(len(h)) / room.fs
+    I = h > 0
+
+    # The power of the impulse response in dB
+    power = 20 * np.log10(np.abs(h[I]))
+    T = T[I]
+
+    # Adjust to location and amplitude of peak power
+    i_max = np.argmax(power)
+    power -= power[i_max]
+    power = power[i_max:]
+    T = T[i_max:]
+
+    # linear fit
+    a, b = np.polyfit(T, power, 1)
+    est_rt60 = - 60 / a
+
+    if plot:
+        # show result
+        plt.plot(T, power)
+        plt.plot(T, a * T + b, '--')
+        plt.plot(T, np.ones_like(T) * -60, '--')
+        plt.vlines(est_rt60, np.min(power), 0, linestyles='dashed')
+
+        if rt60_tgt is not None:
+            plt.vlines(rt60_tgt, np.min(power), 0)
+
+        plt.legend(['RIR', 'Lin Fit', '-60 dB', 'Estimated T60', 'Target T60',])
+
+        plt.show()
+
+    return est_rt60
+
+
+# Define room dimensions and RT60
+room_dim = [10, 7.5, 3.6]
+rt60 = 0.3
+c = 343
+
+# Create the room and place equipment in it
+room = pra.ShoeBox(
+        room_dim,
+        fs=16000,
+        rt60=rt60,
+        )
+room.add_source([2.2, 3.3, 1.75])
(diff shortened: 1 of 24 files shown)


fakufaku:
Ideally, I would like to add some unit testing, but I'm not sure the 3D case gives very good results now. For `rt60=0.3` it is good, but as the longer the `rt60` the more off it gets.

examples/room_rt60.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # linear fit
+    a, b = np.polyfit(T, power, 1)
+    est_rt60 = - 60 / a
+
+    if plot:
+        # show result
+        plt.plot(T, power)
+        plt.plot(T, a * T + b, '--')
+        plt.plot(T, np.ones_like(T) * -60, '--')
```
ebezzam: I'm wondering if here's better to add the offset to the `-60` line? Now it's like this:

original

But I'm thinking something like this is more representative of a `-60 dB` **drop** rather than `-60 dB`.

offset

ebezzam: with the offset that you computed for the orange line

pyroomacoustics/room.py
```diff
@@ -1164,4 +1204,55 @@ def extrude(self, height):
...
+        # diamond. This is what we are doing here.
+        R = []
+        for l1, l2 in  itertools.combinations(room_dim, 2):
+            R.append(l1 * l2 / np.sqrt(l1 ** 2 + l2 ** 2))
+
+        V = np.prod(room_dim)  # volume
+        # surface computation is diff for 2D and 3D
+        if len(room_dim) == 2:
+            S = np.sum(room_dim)
+            sab_coef = 5  # The Sabine's coefficient needs to be adjusted in 2D
```
ebezzam: For the 2D case, I've trying to get an intuitive understanding of what these coefficients/constrants should be. Here's what I've been able to dig up so far.

Video tutorial (skipping to the constant part): https://www.youtube.com/watch?v=A8tWMnr5pps&feature=youtu.be&t=11m22s

Written up derivation of Sabine: https://courses.physics.illinois.edu/phys406/sp2017/Lecture_Notes/P406POM_Lecture_Notes/Derivation_of_the_Sabine_Equation.pdf

So in the well-known [3D case](https://en.wikipedia.org/wiki/Reverberation#Sabine_equation) there's this constant factor of `24` which seems to come from a factor `4` which comes computing the surface area of the sphere multiplied by a factor `6` dropping down from `ln(10^-6)` for the `-60` drop.

So for 2D, I think the combined factor should be `2*6=12` where `2` comes from the area of a circle and the `6` still comes from the `-60 dB` drop.

pyroomacoustics/room.py
```diff
@@ -1164,4 +1204,55 @@ def extrude(self, height):
...
+        # One possibility is to find the largest sphere (or circle in 2D) that fits in the
+        # diamond. This is what we are doing here.
+        R = []
+        for l1, l2 in  itertools.combinations(room_dim, 2):
+            R.append(l1 * l2 / np.sqrt(l1 ** 2 + l2 ** 2))
+
+        V = np.prod(room_dim)  # volume
+        # surface computation is diff for 2D and 3D
+        if len(room_dim) == 2:
+            S = np.sum(room_dim)
```
ebezzam: Here I think it should be `2*np.sum(room_dim)` to sum the 2D room boundaries. This adds up just two of the walls.

pyroomacoustics/room.py
fakufaku: Amazing! Thanks for doing the digging! That falls square in the range that seems correct. Ampirically, I had this 5, but after correcting the computation of the perimeter, this should become 10. That is quite close to the correct 12 that you derived. I'll make the change.

examples/room_rt60.py
fakufaku: Hum, this is an annoying issue. It stems from not having a proper method for computing the empirical T60. I think either is fine now, so I'll go with the more correct version you propose. The correct method will come in a subsequent pull request.

pyroomacoustics/room.py
fakufaku: Right!

fakufaku:
@ebezzam I have implemented a routine to measure the RT60 automatically in `pyroomacoustics.experimental.measure_rt60`. I have also corrected the computation of absorbtion factor according to your feedback. However, the generated IR and the target RT60 do not match. Especially when the reverberation time is long. This is a bit strange. Maybe we are missing something somewhere ?

fakufaku:
Somewhat resolved by the addition of [inverse Sabine function](https://pyroomacoustics.readthedocs.io/en/pypi-release/pyroomacoustics.acoustics.html#pyroomacoustics.acoustics.inverse_sabine) in PR #75 .

PR #52, opened 2018-09-24
Fix bug in one shot STFT so that it works with single channel input.

ebezzam:
There was no value for `channels` if the input was mono, so `pyroomacoustics.transform.stft.analysis` (and `synthesis`) would error out. I also added a couple unit tests.


fakufaku:
@ebezzam This has been merged already, but in the future please also add a note under `bugfix` in the CHANGELOG.

PR #53, opened 2018-09-24
Add ILRMA and integrate into BSS offline and live examples

jazcarretao:
- Add ILRMA algorithm to BSS methods.
- Modify BSS examples to choose BSS method (AuxIVA or ILRMA) with an input argument
- Modify BSS live examples to choose BSS method (AuxIVA or ILRMA) with an input argument
- Update documentation for BSS

### DIFF
--- a/examples/bss_example.py
+++ b/examples/bss_example.py
@@ -1,23 +1,37 @@
 '''
-Blind Source Separation with Indpendent Vector Analysis
+Blind Source Separation offline example
 =======================================================
 
-Demonstrate how to do blind source separation (BSS) using the indpendent vector
-analysis technique. The method implemented is described in the following
-publication.
+Demonstrate the performance of different blind source separation (BSS) algorithms:
+
+1) Independent Vector Analysis (IVA)
+The method implemented is described in the following publication.
 
     N. Ono, *Stable and fast update rules for independent vector analysis based
     on auxiliary function technique*, Proc. IEEE, WASPAA, 2011.
 
-It works in the STFT domain. The test files were extracted from the
+2) Independent Low-Rank Matrix Analysis (ILRMA)
+The method implemented is described in the following publications
+
+    D. Kitamura, N. Ono, H. Sawada, H. Kameoka, H. Saruwatari, *Determined blind
+    source separation unifying independent vector analysis and nonnegative matrix
+    factorization,* IEEE/ACM Trans. ASLP, vol. 24, no. 9, pp. 1626-1641, September 2016
+
+    D. Kitamura, N. Ono, H. Sawada, H. Kameoka, and H. Saruwatari *Determined Blind
+    Source Separation with Independent Low-Rank Matrix Analysis*, in Audio Source Separation,
+    S. Makino, Ed. Springer, 2018, pp.  125-156.
+
+Both algorithms work in the STFT domain. The test files were extracted from the
 `CMU ARCTIC <http://www.festvox.org/cmu_arctic/>`_ corpus.
 
-Running this script will do two things.
+Depending on the input arguments running this script will do these actions:.
 
-1. It will separate the sources.
+1. Separate the sources.
 2. Show a plot of the clean and separated spectrograms
 3. Show a plot of the SDR and SIR as a function of the number of iterations.
 4. Create a `play(ch)` function that can be used to play the `ch` source (if you are in ipython say).
+5. Save the separated sources as .wav files
+6. Show a GUI where a mixed signals and the separated sources can be played
 
 This script requires the `mir_eval` to run, and `tkinter` and `sounddevice` packages for the GUI option.
 '''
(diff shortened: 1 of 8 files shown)


fakufaku:
Hi Juan, nice pull requests. Will you later add the unit test file ? I did not see it here.

jazcarretao:
Hey Robin,
Yes. It is not there yet. I am planning adding it later.

examples/bss_example.py
```diff
@@ -40,14 +54,15 @@
 if __name__ == '__main__':
 
     import argparse
-    parser = argparse.ArgumentParser(description='Demonstration of blind source separation using IVA.')
+    parser = argparse.ArgumentParser(description='Demonstration of blind source separation using IVA or ILRMA.')
     parser.add_argument('-b', '--block', type=int, default=2048,
             help='STFT block size')
+    parser.add_argument('-a', '--algo', type=str, default='ilrma',
```
fakufaku: Here you can use the `choices` keyword that takes a list as argument to restrict the value of the argument to a few pre-defined values, e.g. `choices=['auxiva', 'ilrma']`. This will prevent an error below if the user provides a non-existent method.

examples/bss_example.py
```diff
@@ -129,14 +143,20 @@ def convergence_callback(Y):
...
-    # shape == (n_chan, n_frames, n_freq)
     X = np.array([pra.stft(ch, L, L, transform=np.fft.rfft, zp_front=L//2, zp_back=L//2) for ch in mics_signals])
     X = np.moveaxis(X, 0, 2)
 
-    # Run AuxIVA
-    Y = pra.bss.auxiva(X, n_iter=30, proj_back=True, callback=convergence_callback)
-
-    # run iSTFT
+    # Run BSS
+    bss_type = args.algo
```
fakufaku: An error will occur if the user provides a parameter that is neither 'auxiva' nor 'ilrma'. Restricting at the input (as suggested above) will solve this.

examples/bss_live.py
```diff
@@ -39,7 +54,9 @@
     import argparse
     parser = argparse.ArgumentParser(description='Records a segment of speech and then performs separation')
     parser.add_argument('-b', '--block', type=int, default=2048,
-            help='STFT block length')
+            help='STFT block size')
+    parser.add_argument('-a', '--algo', type=str, default='auxiva',
```
fakufaku: Same as above concerning `choices` keyword.

examples/bss_live.py
```diff
@@ -80,8 +97,15 @@ def cb_print(*args):
         print('  AuxIVA Iter', it)
         it += 10
 
-    # Run AuxIVA
-    Y = pra.bss.auxiva(X, n_iter=args.n_iter, proj_back=True, callback=cb_print)
+    # Run BSS
```
fakufaku: Same here as above.

pyroomacoustics/bss/ilrma.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # initialize the demixing matrices
+    # The demixing matrix has the following dimensions (nfrequencies, nchannels, nsources),
+    if W0 is None:
+        W = np.array([np.eye(n_chan, n_src) for f in range(n_freq)], dtype=X.dtype)
+    else:
+        W = W0.copy()
+
+    # initialize the nonnegative matrixes with random values
+    T = abs(np.array(np.random.rand(n_freq, n_components, n_src)))
```
fakufaku: I think it should be possible to initialize the `T` and `V` matrices so that they have a power profile matching the input spectrogram. This might give faster convergence of the algorithm.
```
S  # input power spectrogram
P_freq = S.mean(axis=0) / n_src
P_time = S.mean(axis=1) / n_src

T_init = np.random.rand(n_freq, n_components, n_src) * P_freq[:,None,None]
V_init = np.random.rand(n_components, n_frames, n_src) * P_time[None,:,None]
```
Also, I think `rand` gives a value in `[0,1]`. Why is `abs` needed here ?

pyroomacoustics/bss/ilrma.py
```diff
@@ -0,0 +1,146 @@
...
+    callback: func
+        A callback function called every 10 iterations, allows to monitor convergence
+
+    Returns
+    -------
+    Returns an (nframes, nfrequencies, nsources) array. Also returns
+    the demixing matrix W (nfrequencies, nchannels, nsources)
+    if ``return_values`` keyword is True.
+    '''
+    X = np.transpose(X, (1, 0, 2))  # now X is (nfrequencies, nframes, nchannels)
```
fakufaku: It might be cleaner to just rewrite the update equations so that there is no need for the transpose. What do you think ? Is it a lot of work ?

fakufaku: What I would suggest to do is to first write the unit test while we are pretty sure the algorithm is working. Then do the modification to avoid the transpose, and use the unit test to make sure everything works as expected.

examples/bss_live.py
fakufaku: Fixed

examples/bss_example.py
fakufaku: Fixed

PR #54, opened 2018-09-25
Adding check for correct absorption coefficients array size for 2D rooms.

CyrilCadoux:
Originally the error was only raised for 3D rooms.


ebezzam:
@fakufaku, @CyrilCadoux noticed this small problem when digging deeper into the RIR generation part of the package

fakufaku:
@CyrilCadoux Thanks for spotting that! I think we should write more unit tests for the Room class! Just construct a bunch of rooms that span the different options so that we can catch these more easily.

fakufaku:
@CyrilCadoux  This has been merged already, but in the future please also add a note under `bugfix` in the CHANGELOG. I didn't check when I merged, so I added the note now.

ebezzam:
@fakufaku thanks! sorry I forgot to mention the CHANGELOG to Cyril and also forgot it in my PR.

Issue #19, opened 2018-02-21
Make matplotlib imports optional throughout the library


fakufaku:
When matplotlib is loaded on a system without a display (i.e. servers or computing nodes), calling matplotlib with a different backen than 'Agg' will cause a crash.

The solution is to limit matplotlib imports to calls that actually plot something.

boeddeker:
In https://github.com/ipython/ipython/pull/10974 is a PR to fix that and a workaround to hide this bug.

fakufaku:
@boeddeker Thanks for the pointer! I guess implementing the workaround would take the same effort as moving the imports into the dedicated functions...

fakufaku:
It is good to know that it is a known issue of ipython. I spent quite some frustrating time trying to debug that...
Now I just need to sit down and do this ;)

boeddeker:
I also invested some time to figure out where the cause of the bug is.
Only ipython has this bug (Python works fine with matplotlib and Qt backend).

fakufaku:
In my case, I use ipyparallel for parallel simulations and this was particularly frustrating to debug... :-/

PR #55, opened 2018-09-28
Less dependencies

fakufaku:
It is desirable to reduce the number of dependencies to a minimum.

This pull requests consistently limits the calls to `matplotlib` to the functions where necessary. When matplotlib is not installed, the function will issue a warning and return without causing an error. 

This is particularly helpful on headless servers where imports of matplotlib would sometimes cause problems when no display is available, or matplotlib is simply not installed.

In addition, the dependencies on `joblib` and `requests` where found to be not necessary and the code was edited to remove them completely.

Finally, small bugs in the doc were corrected and a few `plt.tight_layout` added where necessary. The README was also edited to reflect this.

This PR fixes issue #19

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -106,13 +106,47 @@ Quick Install
 
 Install the package with pip::
 
-    $ pip install pyroomacoustics
+    pip install pyroomacoustics
 
-The requirements are::
+A [cookiecutter](https://github.com/fakufaku/cookiecutter-pyroomacoustics-sim)
+is available that generates a working simulation script for a few 2D/3D
+scenarios::
 
-* numpy 
-* scipy 
-* matplotlib
+    # if necessary install cookiecutter
+    pip install cookiecutter
+
+    # create the simulation script
+    cookiecutter gh:fakufaku/cookiecutter-pyroomacoustics-sim
+
+Dependencies
+------------
+
+The minimal dependencies are::
+
+    numpy 
+    scipy>=0.18.0
+    Cython
+
+where ``Cython`` is only needed to benefit from the compiled accelerated simulator.
+The simulator itself has a pure Python counterpart, so that this requirement could
+be ignored, but is much slower.
+
+On top of that, some functionalities of the package depend on extra packages::
+
+    samplerate   # for resampling signals
+    matplotlib   # to create graphs and plots
+    sounddevice  # to play sound samples
+
+This package is mainly developed under Python 3.5. We try as much as possible to keep
+things compatible with Python 2.7 and run tests and builds under both. However, the tests
+code coverage is not 100% and it might happen that we break some things in Python 2.7 from
+time to time. We apologize in advance for that.
+
+Under Linux and Mac OS, the compiled accelerators require a valid compiler to
+be installed, typically this is GCC. When no compiler is present, the package
+will still install but default to the pure Python implementation which is much
+slower. On Windows, we provide pre-compiled Python Wheels for Python 3.5 and
+3.6.
 
 Example
 -------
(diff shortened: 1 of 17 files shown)


README.rst
```diff
@@ -106,13 +106,47 @@ Quick Install
...
 
-* numpy 
-* scipy 
-* matplotlib
+    # if necessary install cookiecutter
+    pip install cookiecutter
+
+    # create the simulation script
+    cookiecutter gh:fakufaku/cookiecutter-pyroomacoustics-sim
+
```
ebezzam: Do you want to add that to then run the script you can simply do `python <chosen_script_name>.py`?

Also I don't know if this can be prevented but when I run the cookiecutter to create the simulation script, the command line interface doesn't raise an warning/error when I provide a `script_name` if there's already a file with that name. I only get an error after inserting all the other parameters and the cookiecutter tries to create the script:

```
A file with the same name already exists
ERROR: Stopping generation because post_gen_project hook script didn't exit successfully
Hook script failed (exit status: 1)
```

If possible, it could be good to add a condition when entering the `script_name` to see if the file already exists OR when creating the script it uses a new file name, e.g. `<chosen_script_name_001>`, and lets the user know.

pyroomacoustics/doa/tools_fri_doa_plane.py
```diff
@@ -1385,9 +1383,11 @@ def dirac_recon_ri_half_parallel(G, a_ri, K, M, max_ini=100):
     # generate all the random initialisations
     c_ri_half_all = np.random.randn(sz_coef, max_ini)
 
-    res_all = jl.Parallel(n_jobs=thread_num)(
-        jl.delayed(partial_dirac_recon)(c_ri_half_all[:, loop][:, np.newaxis])
-        for loop in range(max_ini))
```
ebezzam: would `joblib` reduce computational time significantly? if so, it might be worth making a variable `joblib_available` like for `matplotlib`

pyroomacoustics/doa/doa.py
```diff
@@ -383,6 +383,11 @@ def polar_plt_dirac(self, azimuth_ref=None, alpha_ref=None, save_fig=False,
             'dirty image' in the case of FRI.
         """
 
+        if not matplotlib_available:
+            import warnings
+            warnings.warn('Matplotlib is required for plotting')
+            return
```
ebezzam: Could be alright, but wondering: why do you raise a warning and return _instead_ of just raising an error?

README.rst
fakufaku: > Do you want to add that to then run the script you can simply do python <chosen_script_name>.py?
I've thought about that. Why not!

I haven't really dug into the internals of cookiecutter, but testing the filename before creation doesn't seem to be out of the box. However, there is an option to save the input in case you run the command again.

pyroomacoustics/doa/tools_fri_doa_plane.py
fakufaku: It was there only for historical reason. It was limited to one thread. This doesn't change performance at all. We tried it in the past and it didn't improve all that much.

pyroomacoustics/doa/doa.py
fakufaku: My thinking is that we don't want the script to fail just because matplotlib is missing. For example when running simulation on a server. This is not a requirement after all :) So raise a warning and silently return.

fakufaku:
@ebezzam One thing I was wondering is if we should keep matplotlib, sounddevice, mir_eval, and samplerate in `requirements.txt`. This file is not a hard requirements (these are in `setup.py`), it is more a list of package required to run things in the repo. So maybe that could be the list of all packages needed to run all the examples. If you approve, I'll add them back there.

ebezzam:
Makes sense. And in the main README you can specify that these packages are just needed for the examples.

PR #57, opened 2018-10-16
Corrects invalid memory access in build_rir cython code

fakufaku:
Corrects invalid memory access in the ``pyroomacoustics.build_rir`` cython accelerator and adds a unit test that checks the cython code output is correct.

Comments:
* I'm not sure how to add docstring to cython extension code
* The doc stays unchanged

### DIFF
--- a/pyroomacoustics/build_rir.pyx
+++ b/pyroomacoustics/build_rir.pyx
@@ -37,7 +37,7 @@ def fast_rir_builder(
     fdl: int
         The length of the fractional delay filter (should be odd)
     lut_gran: int
-        The number of point per unit in the since interpolation table
+        The number of point per unit in the sinc interpolation table
     '''
 
     fdl2 = (fdl - 1) // 2
@@ -45,6 +45,13 @@ def fast_rir_builder(
 
     assert time.shape[0] == visibility.shape[0]
     assert time.shape[0] == alpha.shape[0]
+    assert fdl % 2 == 1
+
+    # check the size of the return array
+    max_sample = ceil(fs * np.max(time)) + fdl2
+    min_sample = floor(fs * np.min(time)) - fdl2
+    assert min_sample >= 0
+    assert max_sample < rir.shape[0]
 
     # create a look-up table of the sinc function and
     # then use linear interpolation
--- a/pyroomacoustics/build_rir.pyx
+++ b/pyroomacoustics/build_rir.pyx
@@ -3,6 +3,8 @@
 import numpy as np
 cimport cython
 
+from libc.math cimport floor, ceil
+
 @cython.boundscheck(False)
 @cython.wraparound(False)
 def fast_rir_builder(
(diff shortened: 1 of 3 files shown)


pyroomacoustics/tests/test_build_rir.py
```diff
@@ -0,0 +1,91 @@
...
+    
+
+if __name__ == '__main__':
+
+    import matplotlib.pyplot as plt
+
+    for t, a, v in zip(times, alphas, visibilities):
+        ir_ref, ir_cython = build_rir(times[0], alphas[0], visibilities[0], fs, fdl)
+
+        print('Error:', np.max(np.abs(ir_ref - ir_cython)))
```
ebezzam: I get an error of `0.060633160336600614` when running the script. Is that expected or too high? I notice `tol=0.01`.

pyroomacoustics/build_rir.pyx
```diff
@@ -41,9 +43,13 @@ def fast_rir_builder(
     fdl2 = (fdl - 1) // 2
     n_times = time.shape[0]
 
-    # create a look-up table of the since function and
+    assert time.shape[0] == visibility.shape[0]
+    assert time.shape[0] == alpha.shape[0]
+
```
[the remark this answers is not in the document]
fakufaku: Good thinking. Will do.

pyroomacoustics/tests/test_build_rir.py
fakufaku: That seems big... I wonder why this is the case... I get `0.001267`. What is your version of python / anaconda. Can you run a few more tests ? Also the continuous integration didn't fail, which means the output is within tolerance there as well...

fakufaku: You can try to provide a larger `lut_gran` (20 is the default). This variable controls the number of points to use in the interpolation.

pyroomacoustics/build_rir.pyx
```diff
@@ -41,9 +43,13 @@ def fast_rir_builder(
     fdl2 = (fdl - 1) // 2
     n_times = time.shape[0]
 
-    # create a look-up table of the since function and
```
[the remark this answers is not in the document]
fakufaku: Yup... 😓

pyroomacoustics/build_rir.pyx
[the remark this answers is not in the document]
fakufaku: Done. Also checked for in the test.

pyroomacoustics/tests/test_build_rir.py
ebezzam: That (rebuild) was indeed the problem! Perhaps a "build from source" section can be added to the main README? I often forget the command (`python setup.py build_ext --inplace`
) and have to look into chat history 🙈 

I can make a separate PR for this if you want.

fakufaku: We could add it to the workflow for the tests too. Feel free to do it! That'd be great!

fakufaku: And thanks! I'll merge into master and make a release this week.

PR #60, opened 2018-11-20
Enhancement refactor libroom extension in cpp

fakufaku:
@ebezzam @CyrilCadoux 

I have refactored the compiled extension into C++. I have also decided to remove the pure python implementation as it is probably too slow to be really usable. Here are a list of reasons to do so.

The PR is not yet ready to be merged. 
* ~~Some tests are still failing.~~ 
* ~~The correct compilation on all the continuous integration servers need to be tested (and probably fixed).~~
* Wheels for all platforms need to be created

The cpp code is in `pyroomacoustics/libroom_src`.

### Pros

* C++ more readable, cleaner object-oriented code
* Using `pybind11` allows to have clean bindings into python. The cpp objects are directly usable from python. The interface between python and the cpp extension is much cleaner than before.
* Using `eigen` for linear algebra simplifies the code, makes it easier to read and hopefully more efficient
* Removing the pure python implementation (which was pretty dirty anyway) simplifies the code base and the number of test cases to run.

### Cons

* Adds dependencies on `pybind11` and `Eigen`
* No pure python implementation means that we must make sure compilation can work on most platforms
* Probably best to create compiled python wheels for all major platforms (mac, win, linux)

### Checks

Happy PR :smiley:


fakufaku:
@CyrilCadoux Did some progress. I fixed the extension and now all the tests are passing. The continuous integration is still not running though. But it should be good enough to start working with the cpp code.

setup.py
```diff
@@ -1,41 +1,137 @@
...
+        Extension(
+            'pyroomacoustics.libroom',
+            [ os.path.join(libroom_src_dir, f)
+                for f in ['libroom.cpp','wall.cpp', 'geometry.cpp', 'room.cpp']],
+            include_dirs=[
+                '.',
+                libroom_src_dir,
+                str(get_pybind_include()),
+                str(get_pybind_include(user=True)),
+                os.path.join(libroom_src_dir, 'ext/eigen'),
```
ebezzam: Would be good to put a note somewhere that the following needs to be run so that the `Eigen` library files are in the proper place:

`git submodule update --init --recursive`

More on this [here](https://stackoverflow.com/questions/10168449/git-update-submodules-recursively).

fakufaku: Definitely! I will add this to the README and doc.

PR #62, opened 2018-11-27
Enhance ray tracing

fakufaku:
Just starting the PR so that we can track changes with respect to my branch.

Happy PR :smiley:


pyroomacoustics/libroom_src/libroom.cpp
```diff
@@ -30,6 +34,10 @@ Room *create_room(py::list _walls, py::list _obstructing_walls, const Eigen::Mat
     room->obstructing_walls.push_back(owall.cast<int>());
 
   room->dim = room->walls[0].dim;
+  
+  // Specific to ray tracing
+  room->mic_pos = _microphones.col(0);
```
fakufaku: Hum, it would be best if the ray tracing can also handle multiple microphones. In that case, you would need to add a loop that checks for intersection with all the microphones within the ray tracer.

pyroomacoustics/libroom_src/libroom.cpp
```diff
@@ -114,5 +131,37 @@ PYBIND11_MODULE(libroom, m) {
...
+		
+	m.def("solve_quad", &solve_quad,
+		"Solves the quadratic system and outputs real roots");
+		
+	m.def("mic_intersection", &mic_intersection,
+		"Computes the intersection point between the ray and the microphone");
+		
+		
+	// Routines for the ray_tracing package
+	m.def("bilbo", &bilbo,"Computes the bilbo");
```
fakufaku: We can remove this if it isn't needed anymore.

pyroomacoustics/libroom_src/ray_tracing.cpp
```diff
@@ -0,0 +1,50 @@
+#include <Eigen/Dense>
```
fakufaku: It doesn't look like this file is used. We can remove it.

pyroomacoustics/libroom_src/ray_tracing.hpp
```diff
@@ -0,0 +1,20 @@
+#ifndef __RAYTRACING_H__
```
fakufaku: We can also remove this file I think.

pyroomacoustics/libroom_src/room.cpp
```diff
@@ -276,3 +276,422 @@ int Room::image_source_shoebox(
...
+    - sound_speed: The speed of sound
+    * 
+    * Return : true if the scattered ray reached the mic, false otw
+	*/
+	
+	bool result = false;
+	
+	Vector1i next_wall_idx(-1);
+	
+	next_wall_hit(hit_point, mic_pos, next_wall_idx);
```
fakufaku: In the case of the scattered ray, the only walls that can intersect with the rays are concave walls. If you think of the case of a convex room, there will never be a wall intersecting the scattered ray, in other word, the microphone is visible from every wall surface. A list of the concave walls indices is maintained in `Room::obstructing walls`. Only these walls need to be checked for intersection.

pyroomacoustics/libroom_src/room.cpp
```diff
@@ -276,3 +276,422 @@ int Room::image_source_shoebox(
...
+	
+	
+	// No wall intersecting the direct scattered ray
+	if (next_wall_idx[0] == -1){
+		
+		VectorXf mic_hit = mic_intersection(hit_point, mic_pos, mic_pos, radius);
```
fakufaku: I think we are taking quite a performance hit from computing the intersection with the receiver surface. This might not be necessary actually. Just using the distance to the closest point to the receiver on the straight line should be good enough.

pyroomacoustics/libroom_src/room.cpp
```diff
@@ -276,3 +276,422 @@ int Room::image_source_shoebox(
...
+		float hop_dist = (hit_point - mic_hit).norm();
+		
+		update_travel_time(travel_time, hop_dist, sound_speed);
+		
+		// Remember that the energy distance attenuation will be 
+		// performed once all the ray arrived.
+		
+		if (travel_time < time_thres){
```
fakufaku: I think it might be better to stop the ray after an energy threshold rather than time. Typically we want to compute the rays up to 60 dB energy decay.

pyroomacoustics/libroom_src/room.cpp
```diff
@@ -276,3 +276,422 @@ int Room::image_source_shoebox(
...
+		distance = (start - hit_point).norm();
+		total_dist += distance;
+		update_travel_time(travel_time, distance, sound_speed);
+		update_energy_wall(energy, wall);
+		
+		//std::cout << "post travel time : " <<  travel_time <<std::endl;
+		//std::cout << "post energy : " <<  energy <<std::endl;
+		
+		// Check if we reach the time threshold for this ray
+		if (travel_time > time_thres){
```
fakufaku: Energy threshold would be preferable.

pyroomacoustics/libroom_src/utility.cpp
```diff
@@ -0,0 +1,390 @@
...
+	
+	if (angle_between(start-hit_point, n) > M_PI_2){
+		n = (-1)*n;
+	}
+	
+	return energy * scat_coef * cos(angle_between(n, mic_pos-hit_point));
+}
+
+
+void append(float travel_time, float energy, std::vector<entry> &output){
```
fakufaku: This function is not needed, the STL vector class already takes care of resizing the allocated memory when necessary.

pyroomacoustics/libroom_src/room.cpp
```diff
@@ -276,3 +276,422 @@ int Room::image_source_shoebox(
...
+	 * OUTPUT 
+	 * - a 2D vector where each entry is a tuple (time, energy)
+	 * reprensenting a ray (scattered or not) reaching the microphone*/
+	 
+	if ( (source_pos-mic_pos).norm() < mic_radius ){
+		std::cerr << "The source is inside the microphone ! " <<std::endl;
+		throw std::exception();
+	}
+	
+	std::vector<entry> output(10000);
```
fakufaku: Here this should be just
```
std::vector<entry> output;
```
The vector will take of resizing when necessary.

pyroomacoustics/libroom_src/tests/test_ray_tracing.py
```diff
@@ -0,0 +1,206 @@
...
+            [2.5, ],
+            [7, ],
+            [1., ],
+        ])
+
+        room = pra.libroom_new.Room(walls, obstructing_walls, microphones)
+
+    return room
+
+def compute_rir(log, time_thres, fs, plot=True):
```
fakufaku: You can use the fast RIR builder here:
```
def compute_rir(log, time_thres, fs, plot=True):


    TIME = 0
    ENERGY = 1

    time = np.array([e[TIME] for e in log], dtype=np.float)
    alpha = np.array([e[ENERGY] for e in log], dtype=np.float)
    visibility = np.ones(time.shape[0], dtype=np.int32)

    # ======= WITH FRACTIONAL PART =======

    # the python utilities to compute the rir
    fdl = pra.constants.get('frac_delay_length')
    fdl2 = (fdl - 1) // 2  # Integer division

    ir = np.zeros(int(time_thres * fs) + fdl)

    from pyroomacoustics import build_rir
    build_rir.fast_rir_builder(ir, time, alpha, visibility, fs, fdl)

    if plot :
        x = np.arange(len(ir)) / fs
        plt.figure()
        plt.plot(x, ir)
        plt.title("RIR")
        plt.show()

    return ir
```

pyroomacoustics/libroom_src/room.cpp
```diff
@@ -276,3 +276,422 @@ int Room::image_source_shoebox(
...
+	// ------------------ INIT --------------------
+	// What we need to trace the ray
+	float energy = 1000;
+	VectorXf start = source_pos;
+	VectorXf end = compute_segment_end(start, max_dist, phi, theta);
+	
+	// The following initializations are arbitrary and does not count since we set
+	// the boolean to false
+	Wall wall = walls[0];
+	Vector1i next_wall_index(0);
```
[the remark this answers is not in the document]
CyrilCadoux: Yes you're right, but I needed to test that this parameter passed by reference was modified correctly.

And using an int passed by reference did not work with Python testing. That's why I used this small (definitively ugly) trick. And a few tests depend on this value so I prefer keeping it.

examples/raytracing.py
```diff
@@ -0,0 +1,1288 @@
...
+
+            if not stop_ray(travel_time, time_thres, energy) :
+                if plot:
+                    draw_segment(start, hit_point)
+                    draw_point(hit_point, markersize=3, color='purple')
+
+                # Ajouter des brackets si on parallelise pas
+                output = output + [[travel_time, energy]]
+
+            break
```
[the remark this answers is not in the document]
CyrilCadoux: Yes, thanks for pointing this out, I removed the break when we talked about this, but I put it again to debug and forgot to remove it again :)

pyroomacoustics/libroom_src/geometry.cpp
```diff
@@ -343,3 +344,91 @@ float area_2d_polygon(const Eigen::MatrixXf &corners)
...
+	 int s2 = v2.size();
+	 
+	 if (s1<2 or s1>3 or s2<2 or s2>3 or s1 != s2){
+		 std::cerr << "size of v1 :" << s1 << std::endl;
+		 std::cerr << "size of v2 :" << s2 << std::endl;
+		 std::cerr << "angle_between : Only 2D and 3D vectors are supported" << std::endl;
+		throw std::exception();
+	 }
+
+	return 	acos(clamp(v1.normalized().dot(v2.normalized()), -1., 1.));
```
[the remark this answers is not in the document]
CyrilCadoux: I think I can indeed use the cosine value, let me change it :)

pyroomacoustics/libroom_src/room.cpp
CyrilCadoux: OK nice, I'll take this into account, I thought the obstructing walls variable was only for ISM. I did not want to touch it.

pyroomacoustics/libroom_src/utility.cpp
CyrilCadoux: As we have a lot of entries, I wrote this function to avoid several useless resizing (ie it is non-sense to have a vector with an initial capacity of 2 when we can directly set it higher right ?)

So if I remove this function, there will be more automatic calls to the resize function and thus more copies of the data, is that really something we want ?

pyroomacoustics/libroom_src/room.cpp
CyrilCadoux: When we construct the room, what should the obstructing_walls list contain ? 
Wall objects that are then converted to int through strange casting ? Or just indices of the walls that can be obstructing ? 

Just to know how I should modify this function when we call it for a scattering ray

pyroomacoustics/libroom_src/room.cpp
CyrilCadoux: The problem is that with the ray tracing, some parts of the RIR are composed of a lot of low-energy entries. 
So setting an energy threshold will not only shorten the tail, but it will also reduce the amplitude of some parts of the rir. Do I have to worry about this then ? Or is it juste a collateral damage ?

pyroomacoustics/libroom_src/room.cpp
fakufaku: The indices of the walls. They are compute on the phone side

fakufaku: Python side!

examples/raytracing.py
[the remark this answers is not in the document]
CyrilCadoux: In fact, the python implementation is late of a few changes with respect to the cpp implementation. Is it worth it that I update this python implementation ? Or will it be removed soon ?

pyroomacoustics/libroom_src/room.cpp
```diff
@@ -276,3 +276,422 @@ int Room::image_source_shoebox(
...
+		if (travel_time > time_thres){
+			break;
+		}
+		
+		// Let's simulate the scattered ray induced by the rebound on
+		// the wall
+		
+		if (scatter_coef > 0){
+			
+			float ray_scat_energy = compute_scat_energy(energy,
```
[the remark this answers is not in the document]
CyrilCadoux: Are you talking about the indentation of the parameters (which I tried to align with the first one) ? Or are you talking about the overall indentation ? 

I have the habit to indent everything using 1 tab

pyroomacoustics/libroom_src/room.cpp
CyrilCadoux: What do you mean by "_using the distance to the closest point to the receiver on the straight line_" ?

I could indeed trace a segment [wall_hit, mic_center] and then take the point that is  'mic_radius' away from the mic_center.

Was that what you meant ?

examples/raytracing.py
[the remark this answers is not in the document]
fakufaku: No, there's no need to maintain the python draft. I had commented before you wrote the cpp part :) Sorry for that.

pyroomacoustics/libroom_src/room.cpp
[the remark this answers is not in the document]
fakufaku: I see. The project is using 2 spaces for the cpp code. Four spaces in python. I am sorry to go against what you usually do but it might be better to keep the indentation consistent. This can probably also fixed automatically in some way.

pyroomacoustics/libroom_src/geometry.cpp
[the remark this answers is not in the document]
fakufaku: Thanks!

pyroomacoustics/libroom_src/room.cpp
fakufaku: No need to do that. I'm thinking that we can use directly the same computations that check for the intersection. This is related to the above comment that describes how to compute the distance from line to point.

1) Compute the distance from ray (line) to microphone (point)
2) If the distance is less than the receiver radius, there is an intersection
3) Use the point closest to the microphone on the ray to compute the distance. In fact, the traveled distance is already computed as a by-product of the intersection check.

I know that this makes the distance 'inexact', but if you think about it, the ratio of the error to the traveled distance is in fact quite small in most cases. This should be good enough for us. We need to keep in mind that the ray tracing process itself is an approximation of the true propagation of sound in space.

pyroomacoustics/libroom_src/room.cpp
fakufaku: Ok, you are right that we are interested in the final energy of the RIR itself, so setting the ray minimum energy to 60 dB might result in not enough energy in the tail. However, the time threshold has similar problems.

Maybe we can specify both thresholds. For rays we will indeed need to specify something smaller than 60 dB. Maybe 90 or 100 dB. The time threshold is useful if we want to limit computations.

pyroomacoustics/libroom_src/utility.cpp
fakufaku: The `vector` class has already constant amortized cost for the `push_back` function. Internally it does exactly what you are doing. When the vector is too small, it will double the size of the allocated storage (there is typically more storage allocated than the current size of the vector).

From [this reference](http://www.cplusplus.com/reference/vector/vector/):
```
Internally, vectors use a dynamically allocated array to store their elements. This array may need to be reallocated in order to grow in size when new elements are inserted, which implies allocating a new array and moving all elements to it. This is a relatively expensive task in terms of processing time, and thus, vectors do not reallocate each time an element is added to the container.
```

It is indeed not a bad idea to pre-allocate some large chunk of memory, but the way you are currently doing it is also update the `size` and filling with zeros. Instead you should use the `reserve` function to only change the size of the allocated memory (not of the vector).

In fact vectors use contiguous storage, which might not be needed in our case. It might be more efficient to use a [`forward_list`](http://www.cplusplus.com/reference/forward_list/forward_list/).

fakufaku:
@CyrilCadoux One more comment. For the scattered ray, you need to make sure the receiver is visible the point of impact on the wall. For obstructing walls, if the receiver is on the back of the wall, no scattered ray should go back to it.

pyroomacoustics/libroom_src/tests/test_ray_tracing.py
CyrilCadoux: I have a problem with this function. 
The time parameter is defined as follows : 

`time: ndarray (double)
        The array of delays for the image sources`

And then there is an assertion to check that `floor(fs * np.min(time)) - fdl2 >= 0`

But for raytracing, min(time) is 0, so this statement is always wrong :/

pyroomacoustics/libroom_src/room.cpp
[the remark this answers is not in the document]
CyrilCadoux: Ok it works. Done.

pyroomacoustics/libroom_src/room.hpp
```diff
@@ -5,8 +5,10 @@
 #include <stack>
 #include <Eigen/Dense>
 #include "wall.hpp"
+#include "utility.hpp"
 
 typedef Eigen::Matrix<int, Eigen::Dynamic, 1> VectorXi;
+typedef Eigen::Matrix<int, 1, 1> Vector1i;
```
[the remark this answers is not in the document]
CyrilCadoux: Removed.

pyroomacoustics/libroom_src/geometry.cpp
```diff
@@ -343,3 +344,91 @@ float area_2d_polygon(const Eigen::MatrixXf &corners)
...
+	 
+	 if (s_start != s_point or s_start != s_end or s_end != s_point){
+		std::cerr << "The 3 vectors objects must have the same dimension !" << std::endl;
+		std::cerr << "Dim start = " << s_start <<std::endl;
+		std::cerr << "Dim end = " << s_end <<std::endl;
+		std::cerr << "Dim point = " << s_point <<std::endl;
+		throw std::exception();
+	 }
+	 
+	 if (s_start == 2){
```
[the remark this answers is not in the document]
CyrilCadoux: Done.

pyroomacoustics/libroom_src/libroom.cpp
CyrilCadoux: Removed

pyroomacoustics/libroom_src/ray_tracing.cpp
CyrilCadoux: Done.

pyroomacoustics/libroom_src/ray_tracing.hpp
CyrilCadoux: Done.

pyroomacoustics/libroom_src/room.cpp
CyrilCadoux: I see your point and in fact this is already what I do in the intersects_mic function.

But the 'distance' variable should store the distance between the last wall_hit and the precise mic_hit. 

So for me, if we want to use an approximation, it would make more sense to  simply take the distance between the last wall_hit and the microphone center.

(Also for the precise code that you quoted, the scattered ray is shooted directly towards the microphone center. So, in that case I can obviously simplify my formula for the distance and juste obtain it like this : `distance = (hit_point - mic_pos).norm() - radius`)

pyroomacoustics/libroom_src/utility.cpp
CyrilCadoux: Looking at [this page](https://github.com/pybind/pybind11/blob/master/include/pybind11/stl.h), it seems that forward lists are not supported by pybind11

Do you think it would be worth to try with basic (double linked) lists ?

pyroomacoustics/libroom_src/room.cpp
CyrilCadoux: Done.

pyroomacoustics/libroom_src/room.cpp
CyrilCadoux: I am a little bit confused with this notion of dB for the energy. Do I have to set a concrete threshold and put it as a parameter (as for the time threshold ) ? Or do I need to compute the energy threshold directly from the initial energy ?

pyroomacoustics/libroom_src/tests/test_ray_tracing.py
fakufaku: Yes, that's a problem: the Ray tracer should not give minimum time of zero. This would only happen if the source were exactly at the same location as the mic. This problem comes from the output vector having empty entries when you create it. I have tried this part and I could solve it by setting the size of output vector to zero at the beginning the algorithm.

pyroomacoustics/libroom_src/room.cpp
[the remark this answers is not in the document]
fakufaku: Fantastic.

pyroomacoustics/libroom_src/room.cpp
fakufaku: Right, then let's do that for scattered rays.

For the other rays, here again we are not trying to find the exact timing, but more the overall energy that hits the receiver in some space/time slot. A different of a few tens of centimeters will not change the overall result.

pyroomacoustics/libroom_src/room.cpp
fakufaku: Sorry, the decay is compared to the start energy. So when going to dB, you would compute something like
```
E_decay_db = 10 * log10(E_0 / E_hit)
```
where `E_hit` is the remaining energy of the ray hitting the receiver. Of course the computation do not need to be in dB. A value of 60 dB corresponds to attenuation of the energy by a factor `1e6`. In general, we want to stop the RIR computation when the (accumulated) energy drops below this. As you correctly pointed out, we thus need to set the threshold for a single ray lower than that. We can experiment with `1e9` or `1e10` (90 and 100 dB, resp.).

pyroomacoustics/libroom_src/utility.cpp
fakufaku: Yes, doubly-linked list consume a little bit more memory (one extra pointer), but I don't think this should be a problem. Let's try with this.

fakufaku:
@CyrilCadoux Congrats, I just merged the code into `next_gen_simulator`, the branch where I will continue to develop the new ray tracing simulator. Thanks for all the good work!

PR #73, opened 2019-01-07
Subspace noise removal

ebezzam:
This PR includes work from [this mini-project](https://github.com/GimenaSegrelles/Audio-and-Acoustics-Signal-Processing-Speech-Enhancement-Project) by @GimenaSegrelles and Mathieu Lecoq for the EPFL audio course.

---

Happy PR :smiley:

I ran into an error when running `nosetests` on the DFT unit tests... @fakufaku have you had this?
```
malloc: *** error for object 0x7fddad8f9748: incorrect checksum for freed object - object was probably modified after being freed.
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Added
 - Adds several options to ``pyroomacoustics.room.Room.simulate`` to finely
   control the SNR of the microphone signals and also return the microphone
   signals with individual sources, prior to mix (usefull for BSS evaluation)
+- Add subspace denoising approach in ``pyroomacoustics.denoise.subspace``.
 
 Changed
 ~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -109,7 +109,7 @@ Install the package with pip::
 
     pip install pyroomacoustics
 
-A [cookiecutter](https://github.com/fakufaku/cookiecutter-pyroomacoustics-sim)
+A `cookiecutter <https://github.com/fakufaku/cookiecutter-pyroomacoustics-sim>`_
 is available that generates a working simulation script for a few 2D/3D
 scenarios::
 
@@ -187,7 +187,7 @@ beamformer in a room.
     plt.show()
 
 A comprehensive set of examples covering most of the functionalities
-of the package can be found in the ``examples`` folder of the `github
+of the package can be found in the ``examples`` folder of the `GitHub
 repository <https://github.com/LCAV/pyroomacoustics/tree/master/examples>`_.
 
 Authors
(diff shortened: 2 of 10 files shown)


fakufaku:
Pretty cool! I didn't know about the subspace denoising method. I will look at it!

I have been having some issues with nosetests depending on the conda environment used. Sometimes the environment gets corrupted and I get weird errors. I would try to create a fresh environment with only the minimal dependencies of the pyroomacoustics test suite and try running the tests again.

One thing that happen sometimes is that I use a conda env that doesn't have nosetests installed. But if nosetests is installed in the root environment, then the root nosetests is run within the test env, which creates problems. If nosetests is not installed in the test env, installing it might make the problem go away.

Before merging this commit, could you please make sure all the code is properly attributed to the authors (@GimenaSegrelles and Mathieu Lecoq, and yourself where necessary) ? I think we need to add the attribution in comments at the top of the actual file. In addition, I think it would be nice to keep a file with the names of the contributors at the root of the repo. At the moment, it is a bit disparate. We have some names in the files, some names in the README (which should just be the core developers), and some commit information, which is only in git. In this PR, just do this for the added code. We can revisit the rest later.

ebezzam:
Yeah it's an interesting approach working in the time domain!

For the author attributes, you mean like at the top of [this file](https://github.com/LCAV/pyroomacoustics/blob/master/pyroomacoustics/bss/auxiva.py)?

Thanks for the nosetests tip, I'll try this out 🙂

ebezzam:
I did something like the top answer on [this post](https://softwareengineering.stackexchange.com/questions/125836/do-you-have-to-include-a-license-notice-with-every-source-file), using MIT license instead.

fakufaku:
Thanks! But... 🙈 I am just thinking that if we do it this way the boring license stuff will show up on every page of the doc.

It seems a solution is to put the license info *before* the docstring in lines commented with `#` (https://stackoverflow.com/questions/17209712/how-do-i-get-sphinx-to-ignore-the-gpl-notice-in-my-docstring)

```
# Boring
# License
# stuff (c)
'''
The docstring starts here

...
'''
```

ebezzam:
Oops didn't think about it popping up in the docs!

fakufaku:
Thanks, did you check this doesn't show up in a local build of the doc ?

ebezzam:
Yes, it doesn't show up with `#`. I also tried building the doc with the `""" ..."""` comment style and the "boring license stuff" indeed showed up.

fakufaku:
It seems the only mention is as `H. L. Van Trees` in the current code, so all is good. I'll merge this.

fakufaku:
Done, thanks @ebezzam !

ebezzam:
Thanks @fakufaku, and to @GimenaSegrelles and Mathieu Lecoq for their work during the semester!

PR #74, opened 2019-01-11
Add iterative Wiener filtering denoising.

ebezzam:
This PR includes work from [this mini-project](https://github.com/lcolbois/aasp-noise-red) by @lcolbois and @liolioclaudio for the EPFL audio course.

---

Happy PR :smiley:

### DIFF
--- a/pyroomacoustics/utilities.py
+++ b/pyroomacoustics/utilities.py
@@ -27,6 +27,7 @@ from scipy import signal
 from scipy.io import wavfile
 
 from .parameters import constants, eps
+from .sync import correlate
 
 
 def create_noisy_signal(signal_fp, snr, noise_fp=None):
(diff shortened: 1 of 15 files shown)


pyroomacoustics/utilities.py
```diff
@@ -462,10 +544,64 @@ def levinson(r, b):
...
-        x = np.concatenate((x, np.zeros(1 if len(b.shape) == 1 else (1, b.shape[1]))), axis=0) + q*np.conj(a[::-1, np.newaxis])
+        if len(b.shape) == 1:
+            x = np.concatenate((x, np.zeros(1))) + q * np.conj(a[::-1])
+        else:
+            x = np.concatenate((x, np.zeros((1, b.shape[1]))), axis=0) + q * np.conj(a[::-1, np.newaxis])
 
     return x
 
+
+def bac(x, p):
```
fakufaku: This implementation is quite inefficient. This can be implemented efficiently  using the FFT.

1. Take FFT with appropriate zero-padding
2. Take squared magnitude
3. Take iFFT
4. Apply appropriate normalization and truncation

fakufaku: I guess the difference in efficiency might depend on the length of x, `L`. If `p < log_2(L)`, then the direct sum might be more efficient (on paper at least, the python loop will be super slow).

fakufaku: Here is an example of the cross-correlation implemented in this way: https://github.com/LCAV/pyroomacoustics/blob/master/pyroomacoustics/sync.py#L45

fakufaku: At least compute the inner loop with `np.inner` or `np.mean`

PR #77, opened 2019-01-22
Add option for number of RIRs to plot.

ebezzam:
Option when plotting RIRs to only plot few channels. Useful when checking RIRs for mic arrays; otherwise, the plot is too cramped.

---

Happy PR :smiley:


ebezzam:
Thanks for the comment, I didn't think about multiple sources! I changed it to settings pairs, but you can also give an int if you want to plot the RIRs from a particular mic to all sources.

pyroomacoustics/room.py
```diff
@@ -702,11 +723,40 @@ def plot(self, img_order=None, freq=None, figsize=None, no_axis=False, mic_marke
 
             return fig, ax
 
-    def plot_rir(self, FD=False):
+    def plot_rir(self, pairs=None, FD=False):
```
fakufaku: I'm not completely satisfied with keyword `pairs` since it does not only describe pairs, but also select a microphone. Maybe we could refactor it as `select` or something like that ?

fakufaku:
Thanks @ebezzam !

PR #83, opened 2019-02-26
Added precision parameter for stft and dft. Updated bss unit test

jazcarretao:

Happy PR :smiley:


CHANGELOG.rst
```diff
@@ -13,6 +13,8 @@ adheres to `Semantic Versioning <http://semver.org/spec/v2.0.0.html>`_.
 
 Added
 ~~~~~
+- Precision parameter for stft
```
fakufaku: Maybe a tad more explanations would be nice, e.g., (float32 vs float64). Does not need to be long.

pyroomacoustics/bss/ilrma.py
```diff
@@ -73,18 +72,18 @@ def ilrma(X, n_src=None, n_iter=20, proj_back=False, W0=None,
     machine_epsilon = np.finfo(float).eps
 
     for n in range(0, n_src):
-        R[:, :, n] = np.dot(T[:,:, n], V[:,:,n])
+        R[:,:,n] = np.dot(T[:,:,n], V[:,:,n])
```
fakufaku: You removed spaces here...

pyroomacoustics/bss/ilrma.py
```diff
@@ -73,18 +72,18 @@ def ilrma(X, n_src=None, n_iter=20, proj_back=False, W0=None,
...
 
     # Compute the demixed output
     def demix(Y, X, W):
         for f in range(n_freq):
-            Y[:,f,:] = np.dot(X[:,f,:], np.conj(W[f,:,:]))
+            Y[:, f, :] = np.dot(X[:,f,:], np.conj(W[f,:,:]))
```
fakufaku: ... and added them here. Let's try to be consistent and remove them everywhere.

pyroomacoustics/bss/tests/test_bss.py
```diff
@@ -16,19 +16,18 @@
...
+# If True test the bss algorithm
+choices = {'auxIVA' : True, 'ILRMA' : True, 'sparseauxIVA' : True}
 
-def test_auxiva():
+# List of frame lengths to test
+L = [256, 512, 1024, 2048, 4096]
 
-    # STFT frame length
-    L = 256
+def test_bss(algo='auxiva', L=256):
```
fakufaku: I see that you have regrouped testing of all algorithms here. This is nice for reduction of code duplication.

The test suite will call all functions that start with `test_` or similar, however, I don't think it will provide argument. Thus, during test, `test_bss()` will be only called once and the other algorithms won't be tested. If you want to run all the tests, you should add short wrapper functions, e.g.,
```
def test_bss_auxiva_256():
    test_bss(algo='auxIVA', L=256)
```
and repeat for all parameters to test.

pyroomacoustics/transform/dft.py
```diff
@@ -1,5 +1,5 @@
 # Author: Eric Bezzam
-# Last modification: July 2, 2018
+# Last modification: February 22, 2019
 
 """
 Class for performing the Dicrete Fourier Transform (DFT) and inverse DFT for
```
fakufaku: `Dicrete` -> `Discrete`
(I know this comes from a previous commit, but let's fix it here 😄 )

pyroomacoustics/transform/dft.py
```diff
@@ -75,38 +74,48 @@ class DFT(object):
...
-        
-        if bits == 32:
-            time_dtype = 'float32'
-            freq_dtype = 'complex64'
-        elif bits == 64:
-            time_dtype = 'float64'
-            freq_dtype = 'complex128'
+
+        if bits is not None and precision is not None:
+            warnings.warn('Deprecated keyword bits ignored in favor of new keyword precision', DeprecationWarning)
```
fakufaku: Maybe add quotation marks around keyword: `bits`, `precision`

pyroomacoustics/transform/stft.py
```diff
@@ -44,7 +43,7 @@ class STFT(object):
         hop size
     analysis_window : numpy array
         window applied to block before analysis
-    synthesis : numpy array
+    synthesis_window : numpy array
```
fakufaku: I agree with renaming the keyword for consistency, but did you check it doesn't break any other code/example ?

pyroomacoustics/transform/stft.py
```diff
@@ -681,7 +697,7 @@ def _overlap_and_add(self, x=None):
 " ---------------------------------------------------------------------------- "
 # Authors: Robin Scheibler, Ivan Dokmanic, Sidney Barthe
 
-def analysis(x, L, hop, win=None, zp_back=0, zp_front=0):
+def analysis(x, L, hop, win=None, zp_back=0, zp_front=0, precision='double'):
```
fakufaku: Here, I think we should not provide the `precision` keyword. Instead, we infer it from `x`'s type. Namely, call `STFT` with `precision=x.dtype` on line 735.

Same comment apply to the `synthesis` function.

pyroomacoustics/transform/stft.py
```diff
@@ -724,7 +745,7 @@ def analysis(x, L, hop, win=None, zp_back=0, zp_front=0):
 
 
 # inverse STFT
-def synthesis(X, L, hop, win=None, zp_back=0, zp_front=0):
+def synthesis(X, L, hop, win=None, zp_back=0, zp_front=0, precision='double'):
```
fakufaku: Same comment as above for `analysis` function.

pyroomacoustics/transform/stft.py
jazcarretao: Within the `__init__` constructor the keyword `synthesis_window` was consistently used already as follows
   ``` 
def __init__(self, N, hop=None, analysis_window=None,
                 synthesis_window=None, channels=1, transform='numpy',
                 streaming=True, precision='double', **kwargs):

        self.synthesis_window = synthesis_window
```

 so renaming only the docstring should not break other parts. Am I right?

File: pyroomacoustics/transform/stft.py [hunk 10]
```diff
@@ -81,6 +86,16 @@ def __init__(self, N, hop=None, analysis_window=None,
         else:
             self.hop = self.num_samples
 
+        if precision == np.float32 or precision == np.complex64 or precision == 'single':
+            self.time_dtype = np.float32
+            self.freq_dtype = np.complex64
+        elif precision == np.float64 or precision == np.complex128 or precision == 'double':
+            self.time_dtype = np.float64
+            self.freq_dtype = np.complex128
+        else:
```
fakufaku: Here I think we should handle a few more cases. For example what if we call `analysis`, which detects the type automatically, on a `np.int16` array ? In the current situation, the call will fail. Ideally, we would like a cast to happen. Without more information, a cast to `np.float64` is desirable.

Most likely, the only time we will use this option is to set it to `single` when needed. The rest of the time, we'd like it set to `double`. So I think it is safe to do
```
if precision == np.float32 or precision == np.complex64 or precision == 'single':
    self.time_dtype = np.float32
    self.freq_dtype = np.complex64
else:
    self.time_dtype = np.float64
    self.freq_dtype = np.complex128
 ```
in other words, if single precision is not requested expressly, we set double precision regardless of the value of the parameter.

What do you think ?

File: pyroomacoustics/transform/stft.py (hunk 7 above)
Comment (fakufaku) [reply pyroomacoustics/transform/stft.py]: Oh, I missed this! Then no problem, nothing should be broken! 😄

File: pyroomacoustics/transform/dft.py [hunk 11]
```diff
@@ -75,38 +74,48 @@ class DFT(object):
...
+        if bits is not None and precision is not None:
+            warnings.warn('Deprecated keyword "bits" ignored in favor of new keyword "precision"', DeprecationWarning)
+        elif bits is not None and precision is None:
+            warnings.warn('Keyword "bits" is deprecated and has been replaced by "precision"')
+            if bits == 32:
+                precision = 'single'
+            elif bits == 64:
+                precision = 'double'
+
+        if precision == np.float32 or precision == np.complex64 or precision == 'single':
```
fakufaku: Maybe here as well we should just set to double whenever single is not explicitely required.

pyroomacoustics/bss/tests/test_bss.py
```diff
@@ -0,0 +1,168 @@
...
+    y_aligned = y[:,L//2:ref.shape[1]+L//2]
+
+    mse = np.mean((ref[:,:y_aligned.shape[1],0] - y_aligned)**2)
+    input_variance = np.var(np.concatenate(signals))
+
+    print('%s with a %d frame length: Relative MSE (expected less than %.e)'
+          % (algo, L, max_mse), mse / input_variance)
+    assert (mse / input_variance) < max_mse
+
+def test_bss_auxiva_256():
```
fakufaku: I was thinking we could just have run all of these in a `for` loop... Sorry for that...

pyroomacoustics/transform/stft.py
jazcarretao: I agree with this point. Then the docstring would look something like this?
```
    precision : string, np.float32, np.float64, np.complex64, np.complex128, optional
        How many precision bits to use for the input.
        If 'single'/np.float32/np.complex64, 32 bits for real inputs or 64 for complex spectrum.
        Otherwise, cast to 64 bits for real inputs or 128 for complex spectrum (default)
```

pyroomacoustics/bss/tests/test_bss.py
jazcarretao: It is okey, I used vim :D
However, it is not clear for me your point. You mean running the test functions in a `for` loop for each frame length (similar to how it was before)?

pyroomacoustics/transform/stft.py
fakufaku: Looks good!

pyroomacoustics/bss/tests/test_bss.py
fakufaku: It just means there is a single test function with a loop inside. Previously it was all in the main, there was no test function.

fakufaku:
Thanks @jazcarretao ! Merged into master.

ebezzam:
Nice work @jazcarretao :)

PR #96, opened 2019-05-27
Fixing some warnings in tests

fakufaku:
This PR fixes a bunch of warnings

* `FutureWarning` when calling `numpy.linalg.lstsq`
* Warning in tests when there are some zeros in `pyroomacoustics.dB`
* Deprecation of `unittest.assertEquals`
* Testing `DFT` with `pyfftw` when `pyfftw` is not available


xaviliz:
Great! Indeed that was my suspect because I harcoded  the filter length definition. Thanks for your dedication!

PR #100, opened 2019-06-11
[BSS] Faster AuxIVA/ILRMA and overdetermined AuxIVA algorithm

fakufaku:
Faster implementation of AuxIVA that also includes OverIVA (more mics than sources). It also comes with a slightly changed API, Laplace and time-varying Gauss statistical models, and two possible initialization schemes.

- [X] Have you checked that the doc builds properly and that any new file has been added to the repo ? How to do that is covered in the [documentation](https://pyroomacoustics.readthedocs.io/en/pypi-release/contributing.html#documentation).

Happy PR :smiley:


examples/bss_example.py
```diff
@@ -168,10 +170,13 @@ def convergence_callback(Y):
         Y = pra.bss.sparseauxiva(X, S, n_iter=30, proj_back=True,
                                  callback=convergence_callback)
 
+    t_end = time.perf_counter()
+    print("Time for BSS: {:.2f}".format(t_end - t_begin))
```
jazcarretao: Here I would just add the time unit

pyroomacoustics/bss/ilrma.py
```diff
@@ -62,72 +89,97 @@ def ilrma(X, n_src=None, n_iter=20, proj_back=False, W0=None,
...
-    R = np.zeros((n_freq, n_frames, n_src))
+    T = np.array(0.1 + 0.9 * np.random.rand(n_src, n_freq, n_components))
+    V = np.array(0.1 + 0.9 * np.random.rand(n_src, n_frames, n_components))
+    R = np.zeros((n_src, n_freq, n_frames))
     I = np.eye(n_src, n_src)
     U = np.zeros((n_freq, n_src, n_chan, n_chan), dtype=X.dtype)
     product = np.zeros((n_freq, n_chan, n_chan), dtype=X.dtype)
     lambda_aux = np.zeros(n_src)
-    machine_epsilon = np.finfo(float).eps
+    machine_epsilon = 1e-15
```
jazcarretao: I think we should call this parameter the same across algorithms, i.e., either `machine_epsilon` or `eps`. For `auxiva`, I would define it in the beginning as well as we do it here.

pyroomacoustics/bss/tests/test_bss.py
```diff
@@ -90,24 +93,32 @@ def freq_bss(algo='auxiva', L=256):
         S = np.sort(np.argpartition(average, -k)[-k:])
         # Run SparseAuxIva
         Y = pra.bss.sparseauxiva(X, S, n_iter=30, proj_back=True)
-        max_mse = 1e-4
+        max_mse = 1e-1
+    elif algo == 'overiva':
+        Y = pra.bss.auxiva(X, n_src=1, n_iter=30, proj_back=True)
+        max_mse = 0.5
+    print(Y.shape)
```
jazcarretao: Is this for debugging purposes or to generate a test log?
For the test log, it could be interesting to get some algorithm speed results.

examples/bss_example.py
```diff
@@ -40,7 +40,7 @@
 
 This script requires the `mir_eval` to run, and `tkinter` and `sounddevice` packages for the GUI option.
 '''
-
+import time
 import numpy as np
 from scipy.io import wavfile
 from pyroomacoustics.transform import STFT
```
jazcarretao: I think this line (46) can be removed

examples/bss_example.py
```diff
@@ -168,10 +170,13 @@ def convergence_callback(Y):
...
+    t_end = time.perf_counter()
+    print("Time for BSS: {:.2f}".format(t_end - t_begin))
     
     ## STFT Synthesis
     y = pra.transform.synthesis(Y, L, L, zp_front=L//2, zp_back=L//2).T
 
+
     ## Compare SDR and SIR
     sdr, sir, sar, perm = bss_eval_sources(ref[:,:y.shape[1]-L//2,0], y[:,L//2:ref.shape[1]+L//2])
     print('SDR:', sdr)
```
jazcarretao: We can add a separation example of the new overiva to see the performance metrics.

pyroomacoustics/bss/auxiva.py
```diff
@@ -1,131 +1,251 @@
...
-                callback(Y)
+                callback(Y_tmp)
 
         # simple loop as a start
         # shape: (n_frames, n_src)
-        r[:,:] = np.sqrt(np.sum(np.abs(Y * np.conj(Y)), axis=1))
+        if model == "laplace":
+            r[:, :] = 2.0 * np.linalg.norm(Y, axis=0)
+        elif model == "gauss":
+            r[:, :] = (np.linalg.norm(Y, axis=0) ** 2) / n_freq
```
jazcarretao: Would it be more clean to define `gauss` and `laplace` as separate functions and call them here? Or creating a dictionary from where we select the desired function based on `model`.

examples/bss_example.py
fakufaku: Good idea.

pyroomacoustics/bss/tests/test_bss.py
fakufaku: leftover debug...

examples/bss_example.py
fakufaku: Hum, I feel this will overcomplicate the script. I'll postpone this to a dedicated example...

pyroomacoustics/bss/auxiva.py
fakufaku: Because the algorithms are otherwise so similar, I feel a keyword argument is the way to go. But I have modified `sparseauxiva` to use the same API.

examples/bss_example.py
fakufaku: Yep, thanks.

pyroomacoustics/bss/ilrma.py
fakufaku: I have standardized to `eps` because `machine_epsilon` has a different meaning (the smallest float that can be represented).

fakufaku:
Thanks @jazcarretao for the review!

PR #108, opened 2019-06-18
Style changes and add/change some checks.

ebezzam:

Happy PR :smiley:

### DIFF
--- a/pyroomacoustics/room.py
+++ b/pyroomacoustics/room.py
@@ -289,6 +289,7 @@ from __future__ import print_function, division
 
 import math
 import numpy as np
+import warnings
 import scipy.spatial as spatial
 from scipy.signal import sosfiltfilt
 from scipy.interpolate import interp1d
@@ -473,7 +474,6 @@ class Room(object):
         if mics is not None:
             self.add_microphone_array(mics)
 
-
     def _var_init(
             self,
             fs,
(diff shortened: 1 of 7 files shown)


pyroomacoustics/room.py
```diff
@@ -1743,7 +1744,7 @@ class ShoeBox(Room):
 
     def __init__(self,
                  p,
-                 fs=8000,
+                 fs=16000,
```
fakufaku: I don't think we should change the defaults. This will break code downstream.

pyroomacoustics/room.py
```diff
@@ -1808,28 +1809,32 @@ def __init__(self,
...
+                        '`absorption` parameter is ignored.',
                         DeprecationWarning,
                         )
 
             if isinstance(materials, Material):
                 materials = dict(zip(self.wall_names, [materials] * n_walls))
-            elif not isinstance(materials, dict):
-                raise ValueError('Materials must be a string, Material object, or dictionary')
+            else:
+                raise ValueError('`materials` must be a `Material` object.')
```
fakufaku: Why not keep the dictionary ? I think it can be good if we want to set different materials for walls.

ebezzam: ah sorry, I thought it was a "leftover" like the string! but now I see later on you were creating a dict for all cases to pass to libroom. I'll put it back and also add some docstring :)

pyroomacoustics/room.py
ebezzam: agreed

PR #113, opened 2019-06-25
Add room generator.

ebezzam:

Happy PR :smiley:

Ran `nosetests` and I got the following error:
```
ERROR: Failure: ImportError (bad magic number in 'pyroomacoustics.c_package': b'\x03\xf3\r\n')
```


ebezzam:
I like the idea of having this separate / more general `RoomGenerator` object, which could be an abstract class that takes as input a source generator and noise generator, with this example here being a subclass for ShoeBox rooms / the approach from [this paper](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46107.pdf). (And another example/class can be made based off [this paper](http://danielpovey.com/files/2017_icassp_reverberation.pdf)).

I'll start off with creating the `random` sub-package and switching to `argparse`, and I'll also think more about how an API can be created with less assumptions.

fakufaku:
I just saw your comment about `bad magic number` after running `nosetests`. This happen when the `.pyc` file is left over from a file that do not exist after switching branch. Here `c_package` relates to the C extension in master. You can manually delete the `pyc` files to get rid of the problem.

ebezzam:
> I just saw your comment about `bad magic number` after running `nosetests`. This happen when the `.pyc` file is left over from a file that do not exist after switching branch. Here `c_package` relates to the C extension in master. You can manually delete the `pyc` files to get rid of the problem.

Yes I found this solution online and deleting the `pyc` files was indeed the fix.

PR #125, opened 2019-09-13
Scattering fix

ebezzam:

Happy PR :smiley:

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -47,11 +47,21 @@ using the package, you will then also need to compile the C++ extension for
 that. On Mac and Linux, GCC is required, while Visual C++ 14.0 is necessary for
 `windows <https://wiki.python.org/moin/WindowsCompilers>`__. 
 
-1. Get the source code
+1. Get the source code. Use recursive close so that Eigen (a sub-module of this
+   repository) is also downloaded.
 
    .. code-block:: shell
 
-       git clone git@github.com:LCAV/pyroomacoustics.git
+       git clone --recursive git@github.com:LCAV/pyroomacoustics.git
+
+   Alternatively, you can clone without the `--recursive` flag and directly
+   install the Eigen library. For macOS, you can find installation instruction
+   here: https://stackoverflow.com/a/35658421. After installation you can
+   create a symbolic link as such:
+
+    .. code-block:: shell
+
+        ln -s PATH_TO_EIGEN pyroomacoustics/libroom_src/ext/eigen/Eigen
 
 2. Install a few pre-requisites
 
@@ -59,15 +69,7 @@ that. On Mac and Linux, GCC is required, while Visual C++ 14.0 is necessary for
 
         pip install numpy Cython pybind11
 
-3. Install Eigen library. On macOS, you can find installation instruction here:
-   https://stackoverflow.com/a/35658421. After installation you can create a
-   symbolic link as such:
-
-    .. code-block:: shell
-
-        ln -s PATH_TO_EIGEN pyroomacoustics/libroom_src/ext/eigen/Eigen
-
-4. Compile locally
+3. Compile locally
 
    .. code-block:: shell
 
@@ -80,7 +82,7 @@ that. On Mac and Linux, GCC is required, while Visual C++ 14.0 is necessary for
 
          MACOSX_DEPLOYMENT_TARGET=10.9 python setup.py build_ext --inplace
 
-5. Update ``$PYTHONPATH`` so that python knows where to find the local package
+4. Update ``$PYTHONPATH`` so that python knows where to find the local package
 
    .. code-block:: shell
 
@@ -90,13 +92,13 @@ that. On Mac and Linux, GCC is required, while Visual C++ 14.0 is necessary for
    For windows, see `this question <https://stackoverflow.com/questions/3701646/how-to-add-to-the-pythonpath-in-windows>`__
    on stackoverflow.
 
-6. Install the dependencies listed in ``requirements.txt``
+5. Install the dependencies listed in ``requirements.txt``
 
    .. code-block:: shell
 
       pip install -r requirements.txt
 
-7. Now fire up ``python`` or ``ipython`` and check that the package can be
+6. Now fire up ``python`` or ``ipython`` and check that the package can be
    imported
 
    .. code-block:: python
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -53,11 +53,25 @@ that. On Mac and Linux, GCC is required, while Visual C++ 14.0 is necessary for
 
        git clone git@github.com:LCAV/pyroomacoustics.git
 
-2. Compile locally
+2. Install a few pre-requisites
+
+    .. code-block:: shell
+
+        pip install numpy Cython pybind11
+
+3. Install Eigen library. On macOS, you can find installation instruction here:
+   https://stackoverflow.com/a/35658421. After installation you can create a
+   symbolic link as such:
+
+    .. code-block:: shell
+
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.rst
```diff
@@ -53,11 +53,25 @@ that. On Mac and Linux, GCC is required, while Visual C++ 14.0 is necessary for
...
        git clone git@github.com:LCAV/pyroomacoustics.git
 
-2. Compile locally
+2. Install a few pre-requisites
+
+    .. code-block:: shell
+
+        pip install numpy Cython pybind11
+
+3. Install Eigen library. On macOS, you can find installation instruction here:
```
fakufaku: Eigen is included in the repository as a sub-module. It should be possible to get it via a recursive clone.

CONTRIBUTING.rst
```diff
@@ -53,11 +53,25 @@ that. On Mac and Linux, GCC is required, while Visual C++ 14.0 is necessary for
...
+    .. code-block:: shell
+
+        ln -s PATH_TO_EIGEN pyroomacoustics/libroom_src/ext/eigen/Eigen
+
+4. Compile locally
 
    .. code-block:: shell
 
-         python setup.py build_ext --install
+         python setup.py build_ext --inplace
```
fakufaku: Good catch! 👍

pyroomacoustics/libroom_src/room.cpp
```diff
@@ -693,13 +693,17 @@ bool Room<D>::scat_ray(
     {
       // As the ray is shot towards the microphone center,
       // the hop dist can be easily computed
-      float hop_dist = (hit_point - mic_pos).norm();
+      Vectorf<D> hit_point_to_mic = mic_pos - hit_point;
```
fakufaku: So the vector direction is pointing towards the wall. I believe this is necessary to get the correct angle computation for the Lambert scattering, because the wall normal is pointing behind the wall in our convention. Is this correct ? If so, we could add the explanation in a comment.

CONTRIBUTING.rst
ebezzam: Ok didn't know about recursive clone! So I installed Eigen and created the symbolic link. The recursive clone approach would be to run the following?
```
git clone --recursive git@github.com:LCAV/pyroomacoustics.git
```

pyroomacoustics/libroom_src/room.cpp
fakufaku: Oups! You are right! Yes, you can look for comments in the code to verify, but I think I set the default to be normals pointing outside the room.

fakufaku: In fact, it might be better to wrap the angle computation in a dedicated method of the wall class so that we don't have to worry about this anymore.

ebezzam: Good idea 👍 

Actually looking at [how the normal](https://github.com/LCAV/pyroomacoustics/blob/cb914279b6568635856743322175e2b3c65aed6c/pyroomacoustics/libroom_src/wall.cpp#L81) is computed I'm not sure if it is always pointing outside the room. From just two corners, how could it be known which direction is in/out? But I guess in a dedicated method, we can make the assumption that the mic is in the room in order to perform the correct angle computation?

pyroomacoustics/libroom_src/room.cpp
```diff
@@ -693,13 +693,18 @@ bool Room<D>::scat_ray(
...
       float travel_dist_at_mic = travel_dist + hop_dist;
 
       // compute the scattered energy reaching the microphone
       float m_sq = mic_radius * mic_radius;
       float h_sq = hop_dist * hop_dist;
-      Eigen::VectorXf scat_trans = wall.scatter * transmitted * 2.f * m_sq / h_sq;
+      float p_hit_equal = 1.f - sqrt(1.f - m_sq / h_sq);
+      // cosine angle should be positive, but could be negative if normal is
+      // facing out of room so we take abs
+      float p_lambert = 2 * abs(wall.cosine_angle(hit_point_to_mic));
```
ebezzam: @fakufaku I added this Wall method for computing the cosine angle. Whether the normal is facing in or out, I think this should work since we can either multiply the normal by -1 to flip its direction if it's facing outward or take the absolute value of the cosine angle.

fakufaku: Great! But could you check wether the normal is ourward pointing ? It is better that the result of `wall.cosine_angle` is well defined. In this particular case, the `abs` is fine though.

fakufaku:
BTW, the test `tests/tests_libroom/test_ray_energy.py` is failing. It was depending on the computation of the scattering, so the tests should be fixed to account for the corrected computation.

pyroomacoustics/libroom_src/room.cpp
ebezzam: Should the `abs` be moved into `wall.cosine_angle`? I suppose we would mainly be concerned with angles inside the room, and I could make a comment that the method computes the cosine angle assuming the normal is pointing towards the same half-space as the provided vector, therefore returning values within [0, 1].

Because from just a wall, I don't think you can tell whether the normal is pointing in or out? as you need info about the other walls or assume that the provided vector is pointing into the room.

ebezzam:
Actually `tests/tests_libroom/test_ray_energy.py` doesn't use scattering, and it was failing in the [original branch](https://travis-ci.org/LCAV/pyroomacoustics/jobs/551393864#L699). But I can take a look. 

But there is a new test that's failing due to the scattering: [`test_scat_ray_ok`](https://travis-ci.org/LCAV/pyroomacoustics/jobs/586459326#L921).

ebezzam:
What I notice is the that the initial energy is set differently. 

[ground-truth](https://github.com/LCAV/pyroomacoustics/blob/cb914279b6568635856743322175e2b3c65aed6c/pyroomacoustics/tests/tests_libroom/test_ray_energy.py#L72) vs. [ray tracing simulation](https://github.com/LCAV/pyroomacoustics/blob/next_gen_simulator/pyroomacoustics/libroom_src/room.cpp#L871), there is a factor 2 difference which I also see in the resulting histograms

ebezzam:
From Eq. 5.54 in Dirk's thesis, it seems that it should be multiplied by 2 (instead of 4) as done for the ground-truth histogram. And do we need this `(1-cos gamma/2)` term?

fakufaku:
Hi @ebezzam, I'm not completely sure where this PR was left at. Did we solve the factor 2 problem ?

fakufaku:
Also, as you mentioned to me, the current code does not use scattering for reflections under the ISM model order, which is not correct. I think the fix is very simple.

In `pyroomacoustics/libroom_src/room.cpp`, line 842, remove the second condition from the `if` clause (`!already_in_ism`).

fakufaku:
BTW, I've fixed a bunch of stuff in `next_gen_simulator`, you might want to sync the branches.

ebezzam:
> Hi @ebezzam, I'm not completely sure where this PR was left at. Did we solve the factor 2 problem ?

hey @fakufaku, sorry forgot to push that commit. do you agree it should be multiplied by 2 instead of 4?

ebezzam:
> Also, as you mentioned to me, the current code does not use scattering for reflections under the ISM model order, which is not correct. I think the fix is very simple.
> 
> In `pyroomacoustics/libroom_src/room.cpp`, line 842, remove the second condition from the `if` clause (`!already_in_ism`).

I think that's one thing that needs to be changed but also attenuate the ISM amplitudes by (1 - scattering coefficient)

fakufaku:
Eric, thanks for the PR. I noticed still a few things, like the `is_hybrid_sim` parameters is not set correctly currently (my fault, always set to `True`, even for pure ISM), which leads to wrong energy for ISM that doesn't use ray tracing. But I'm merging the PR so that we can move forward.

PR #155, opened 2020-02-26
Fix error when setting temperature without humidity in ShoeBox constructor

ebezzam:
`humidity` would be set to `None` from the `ShoeBox` constructor so the following statement would fail when the `Physics` constructor was called (due to setting the `temperature`).

```python
if self.H < 0.0 or self.H > 100:
```

---

Happy PR :smiley:

### DIFF
--- a/pyroomacoustics/parameters.py
+++ b/pyroomacoustics/parameters.py
@@ -143,19 +143,22 @@ class Physics(object):
     temperature: float, optional
         The room temperature
     humidity: float in range (0, 100), optional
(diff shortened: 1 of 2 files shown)


fakufaku:
@ebezzam , thanks for spotting this silly mistake of mine! I have fixed the test in `next_gen_simulator`. Would you mind merging once before I pull it in ?

ebezzam:
@fakufaku merged with `next_gen_simulator` 👍

PR #160, opened 2020-04-29
Add AnechoicRoom

duembgen:
## Summary

This pull request is work in progress. I will notify below as soon as it is ready to be viewed. Happy for any input in the meantime. 

## Checklist

## Left to do
- [ ] Replace by AnechoicRoom wherever it makes sense.


pyroomacoustics/signal.py
```diff
@@ -0,0 +1,73 @@
+# Signal object for simulating source signals. 
+# Copyright (C) 2019  Robin Scheibler, Ivan Dokmanic, Sidney Barthe
```
fakufaku, line 2: This should probably be your name here :)

pyroomacoustics/signal.py
```diff
@@ -0,0 +1,73 @@
...
+from __future__ import print_function
+
+import numpy as np
+
+available_types = {
+    'mono': {'frequency': 'Frequency of mono sound (Hz)', 't_max':'Maximum time (s)'},
+    'random': {'N': 'Number of random samples'}
+}
+
+class Signal(object):
```
fakufaku, line 41: I have previously also thought about using a special class for signals so that we can add properties like sampling rate, etc, to it. However, I never fully proceeded due to difficulty to imagine interaction with numpy and the Signal objects in the future. Do we want the hurdle of converting from signals to numpy array, and then back ?

Regardless, at the time I looked into sub-classing the ndarray class, so that we could retain compatibility with numpy, i.e. feed a Signal to a regular numpy routine. Maybe that could be one way to do it ? Here is the numpy [doc](https://numpy.org/doc/stable/user/basics.subclassing.html). Take a look a the simple example.

pyroomacoustics/room.py
```diff
@@ -1586,9 +1587,38 @@ def __init__(self,
...
 
+    def __str__(self):
+        return f'ShoeBox instance in {self.dim}D with {len(self.walls)} walls'
+
     def extrude(self, height):
         ''' Overload the extrude method from 3D rooms '''
 
         Room.extrude(self, np.array([0., 0., height]))
 
+class InfiniteRoom(Room):
```
fakufaku, line 2208: Such a room is also something I have been thinking about for some time :) I'm not completely sure about the InfiniteRoom name though. I'm not sure that is what a user would immediately think about. I'm more in favor of AnechoicRoom or FreeSpace. But there might be yet a better name. The advantage of AnechoicRoom would be to retain the idea of a limitited space (a room), but without reverb.

examples/doa_algorithms_infiniteroom.py
```diff
@@ -0,0 +1,114 @@
+'''
```
fakufaku, line 1: For that one, no need to create a new example, just replace `ShoeBox` in the original script by the new class. The original uses an anechoic room anyway.

.gitignore
```diff
@@ -10,3 +10,8 @@ docs/_build
 dist/
 local_examples/
 *.ipynb_checkpoints
+# files added after editable install (pip install -e . )
+pyroomacoustics.egg-info/
+pyroomacoustics/build_rir.c
+pyroomacoustics/build_rir.cpython-36m-x86_64-linux-gnu.so
```
fakufaku, line 16: You will need to modify this for every version of python and architechture. Maybe just add *.so

duembgen:
Thank you for reviewing! Great to have input this early on. I changed the target branch to next_gen_simulator in the meantime, and will consider your comments in the following. 

Regarding the Signal class I agree it might be an overkill if it means changing the code in many places. In my use-case I only had to change the way the convolution is computed, but I only looked at a subset of the functionalities of pyroomacoustics for now (DOA algorithms). If the other functionalities imply many more changes I agree it is not worth it. 

What I would find useful then is to have a few helper functions to generate commonly used signals (ndarrays), such as white noise, monochromatic signals, etc. Do you think that would be useful, or does it already exist and I just didn't see it?

pyroomacoustics/signal.py
duembgen: My main reason to add the class is exactly like you say, to add properties to the signal. The ndarray is simply one member of it (I call it `self.samples`), but there is also the sampling rate, the type of the signal, etc. That way we don't loose this information throughout the pipeline. Wherever the ndarray is needed, it would be enough to replace the code by `signal.samples`. 

The numpy sublcass seems like a very interesting alternative to not clog the code, especially if we want to keep compatibility with both ndarray and the Signal class. If you think this could be an option I'd be happy to implement a first example. I'd move this to a different pull request though since then it would be come a bigger things.

pyroomacoustics/signal.py
duembgen: I was not sure how you handle these - whoever edits the file adds their name there?

pyroomacoustics/room.py
duembgen: I agree it might not be the best name! I see a few issues:
- InfiniteRoom: indeed, the user might not understand immediately what this means
- AnechoicRoom: My favorite, however I would not make it an actual room, as in all objects having to be inside it, because that would seem like an unnecessary nuisance to the user? That's where my "InfiniteRoom" idea came from: to have walls at infinite distance, such that there is no echo effect, and one does not have to worry about placing objects inside it.
- FreeSpace: It is the most fitting term imo, however "FreeSpace is not a type of Room" so in that sense, it does not comply with inheritance best practices. If we remove inheritance, I would suggest to use this name.

.gitignore
duembgen: Good point!

examples/doa_algorithms_infiniteroom.py
duembgen: Great, I was going to do that, just kept both examples for now to have a quick way of comparing

pyroomacoustics/signal.py
fakufaku: I think so. If you write the code, you own the copyright to that code. When multiple people edit or modify the file, I'm not comletely sure, but my understanding is that you'll list all the people who contributed to the file.

pyroomacoustics/signal.py
fakufaku: Cool. In the beginning I also thought it would be neat to have extra attributes on the signals (e.g. sampling rate obviously). So if we can do it in a smart way I'm up for it.

For your specific use case, I there might also be alternatives.
1. Create a sub-package with routines pretty much like `numpy.zeros` or `numpy.ones` but for typical signals, e.g. white noise, sweep, sine, etc.
2. If the only goal is to add this signal to a sound source, we could add methods to provide the signal of the sound source. In this case, we might also need to provide a sampling frequency to the sound source to make the interface smoother.
```python
my_source = pra.SoundSource(location, fs=16000)
my_source.play_white_noise(length=1.7, power=0.5, offset=0.1)
other_source = pra.SoundSource(other_location, fs=16000)
other_source.play_sweep(length=3., type="linear", fs_low=100, fs_high=21000)
``` 

I think 1. is simplest and most straightforward, and not incompatible with adding 2. at a later time.

pyroomacoustics/room.py
fakufaku: > AnechoicRoom: My favorite choice, however I would not make it an actual room with walls, with all objects having to be inside it. That would seem like an unnecessary nuisance to the user?

I'm definitely with you on that one. Regardless of the name, we should not require dimensions. If there's a rewrite of the class, I don't think that should be a big problem. I don't know which name would be easiest to understand for users. Maybe we can brainstorm a few more names
* `UnboundedSpace`: makes it explicit there are no walls
* `Space`: simple
* `UnboundedRoom`: far fetched, variation on unbounded space
* `NoReverbRoom`: factual, less technical than anechoic

pyroomacoustics/signal.py
duembgen: Great ideas! I agree point 1 would be a nice first step, because these functions will be useful in the classes written in 2. or 3.. I have removed the Signal subclass from this pull request to keep this pull request about the Anechoic room only, but I will create a new pull request soon and include the history there.

pyroomacoustics/room.py
duembgen: I would vote for UnboundedRoom, NoReverbRoom or AnechoicRoom if we are making a subclass of Room. Another idea I had today, but maybe a bit far-fetched:
- `NoRoom`: would also make explicit that there are no walls, while using the "Room" concept anyways

The other option I see after working a bit more on the code is to have the current `Room` class inherit form a new `Space` baseclass (or a different name of course). `Space` has microphones, soundsources, and physical attributes like the speed of sound, etc. The `Room` subclass adds to this space class the concept of walls. This would require more rewriting though, and I let you decide if you think it is worth it! Personally, I think it would lead to the conceptually cleanest code.

Regarding my usecase, I would then use the `Space` class only

duembgen:
I have created a first version of the free-space simulation. As marked in the code, there are a few places where I am not 100% sure of the best implementation, I will expand address these points in a code review below. For now, comparing the output of `examples/doa_algorithms.py` and `examples/doa_algorithms_infiniteroom.py`, The output is not exactly the same, which I believe it should be. Also, there is an ill-conditioning warning from FRIDA. I am looking into it!

pyroomacoustics/room.py
```diff
@@ -2189,15 +2212,230 @@ def extrude(self, height):
...
+        if self.sigma2_awgn is not None:
+            signals += np.random.normal(0.0, np.sqrt(self.sigma2_awgn), signals.shape)
+
+        # record the signals in the microphones
+        self.mic_array.record(signals, self.fs)
+
+        if return_premix:
+            return premix_signals
+
+    # TODO(FD): not sure if we want the far-field option below?
```
duembgen, line 2402: I have included the option to specify a reference mic, in which case the source is considered far-field and the delays are computed with respect to that reference. I am not sure if this is a desired feature, as it does not properly reflect the real physics of the setup.

pyroomacoustics/room.py
```diff
@@ -2189,15 +2212,230 @@ def extrude(self, height):
...
+            signals += np.random.normal(0.0, np.sqrt(self.sigma2_awgn), signals.shape)
+
+        # record the signals in the microphones
+        self.mic_array.record(signals, self.fs)
+
+        if return_premix:
+            return premix_signals
+
+    # TODO(FD): not sure if we want the far-field option below? 
+    def get_delay(self, source_idx, mic_idx, mic_ref_idx=None):
```
duembgen, line 2403: I have included the option to specify a reference mic, in which case the source is considered far-field and the delays are computed with respect to that reference. I am not sure if this is a desired feature, as it does not properly reflect the real physics of the setup.

pyroomacoustics/room.py
```diff
@@ -2189,15 +2212,230 @@ def extrude(self, height):
...
+                delay = int(source_delay + mic_delay)
+                delays[m, s] = delay
+
+                max_sig_len = max(delay + len(self.sources[s].signal), max_sig_len)
+
+        # TODO(FD) not sure if removing the zero-signal stretch poses problems
+        # for applications other than DOA. 
+        min_delay = delays.min()
+        max_sig_len -= min_delay
+        delays -= min_delay
```
duembgen, line 2360: Here, the question is if we want to remove the time delay until the first signal hits the first mic. For DOA, this offset is not important, but maybe it is for other applications that pyroomacoustics is used for?

pyroomacoustics/room.py
```diff
@@ -2189,15 +2212,230 @@ def extrude(self, height):
...
+        delays = np.zeros((M, S), dtype=np.int)
+        for m in range(M):
+            for s in range(S):
+
+                source_delay = self.sources[s].delay * self.fs
+                mic_delay = self.get_delay(s, m)  * self.fs
+
+                #TODO(FD) use floor or round here? 
+                delay = int(source_delay + mic_delay)
+                delays[m, s] = delay
```
duembgen, line 2352: This is a little detail, but I saw that sometimes the integer delay is calculated using the floor operation, instead of rounding. What would you like to use here?

pyroomacoustics/room.py
```diff
@@ -2189,15 +2212,230 @@ def extrude(self, height):
...
+
+        # compute all delays from M microphones to S sources, and 
+        # maximum signal length
+
+        max_sig_len = 0 
+
+        # TODO(FD) make this faster using broadcasting
+        delays = np.zeros((M, S), dtype=np.int)
+        for m in range(M):
+            for s in range(S):
```
duembgen, line 2345: This section could potentially be done faster using broadcasting, but then the readability would suffer. Using it a few times, it doesn't look like this code takes long at all, so I am not sure if we want to trade performance for readability?

duembgen:
@fakufaku the technical side of a first AnechoicRoom implementation is finished. There is a new test
`test_anechoic_room.py` (run it with python to see some plots, with pytest for testing), and also the `examples/doa_algorithms.py` now gives the same results for shoebox vs anechoic, except for FRIDA which I believe is due to randomness.
In the end, it was easiest to make AnechoicRoom inherit from ShoeBox, it seemed like it requires the least rewriting of code. 
Looking forward to your feedback when you have a moment. In the meantime I'll work on the documentation of the new class and see where in the example code we could replace ShoeBox by AnechoicRoom.

duembgen:
I see that the code does not the Python2.7 tests because the `black` autoformatting tool adds trailing commas on function calls that were split over multiple lines, which Python2.7 does not like. Did you have this issue before @fakufaku and know a good fix for it? For now I am manually adding `#fmt:off` flags around these places

fakufaku:
Oh, its seems indeed that is the case. Is it only after the `*args, **kwargs` ? Anyway, let's keep this version python 2.7 compatible for now. Once this one work, we can drop the support I think. Also, this means no f-strings 😢

fakufaku:
BTW, I tried a simple example, and in my case, black doesn't add the trailing comma after `**kwargs` by default. Do you have a special configuration ? If we can avoid littering the code with `# fmt: off` that would be nice.

pyroomacoustics/room.py
```diff
@@ -2189,15 +2154,64 @@ def extrude(self, height):
...
+
+    def __str__(self):
+        return "AnechoicRoom instance in {}D.".format(self.dim)
+
+    def is_inside(self, p):
+        """ Overloaded function to eliminate testing if objects are "inside" room. """
+        # always return True because we want the walls to have no effect.
+        return True
+
+    def get_bbox(self):
```
fakufaku, line 2199: I guess it depends on how you wrote the plotting function too, but should the bbox be infinite or made so that it contains all the microphones and sources ?

pyroomacoustics/room.py
```diff
@@ -1725,6 +1673,8 @@ def simulate(
             raise ValueError("There are no sound sources in the room.")
         if self.mic_array is None:
             raise ValueError("There is no microphone in the room.")
+        if any([s.signal is None for s in self.sources]):
```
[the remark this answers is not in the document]
duembgen: The reason is that a few lines further down you call 
```
        f = lambda i: len(self.sources[i].signal) + np.floor(
            self.sources[i].delay * self.fs
        )
```
which throws an error if any of the sources[i].signal is None. If you think it is important to have silent sources, then I can simply do the fix there?

pyroomacoustics/room.py
duembgen: That's how it is implemented for now: bbox is infinite by default, and then gets reduced to cover all microphones and sources

pyroomacoustics/room.py
```diff
@@ -2189,15 +2154,64 @@ def extrude(self, height):
...
+
+    Parameters
+    ----------
+    dim: int
+        Dimension of the room (2 or 3).
+    kwargs:
+        All other kwargs are passed on to the :py:obj:`pyroomacoustics.room.Room` initializer.
+    """
+
+    def __init__(self, dim, **kwargs):
```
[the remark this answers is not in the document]
duembgen: Ok, your call! Since it is also the convention in the rest of the package, I guess it makes sense to follow the same convention, for consistency purposes.

duembgen:
> BTW, I tried a simple example, and in my case, black (version 19.10b0) doesn't add the trailing comma after `**kwargs` by default. Do you have a special configuration ? If we can avoid littering the code with `# fmt: off/on` that would be nice.

I use the same version and no special configuration. I am looking into it but worse case I can remove the `#fmt` tags and just not run autoformatting on this file anymore (it is mostly in good shape now anyways)

pyroomacoustics/room.py
```diff
@@ -2189,15 +2154,64 @@ def extrude(self, height):
 
         Room.extrude(self, np.array([0.0, 0.0, height]))
 
-        # update the shoebox dim
```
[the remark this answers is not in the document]
duembgen: Indeed, looks like a mistake! I apologize

duembgen:
A note on the plotting function: Nothing has changed for Room, I only refactored a little bit so that plot_walls is separated from the rest and can be overwritten by the AnechoicRoom class, to do nothing. In the process of doing this, I removed quite some duplicate code which came from the fact that there was a disticntion for 2D vs 3D to call either `plot(x[0], x[1], ...)` or `plot(x[0], x[1], x[2], ...)`. If we use `plot(*x, ...)` then we don't have to do this distinction.

pyroomacoustics/room.py
duembgen: I added an error because if there are neither mics nor sources, the bbox stays infinite and one cannot set infinite bounds in matplotlib.

pyroomacoustics/room.py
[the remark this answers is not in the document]
duembgen: I rewrote this and agree now that this makes the code more readable!

pyroomacoustics/room.py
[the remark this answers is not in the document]
duembgen: I changed the code back to allow for non-specified signals, by only calling f if sources[i].signal is specified. We could through an error if none of them are specified though? Then, computing RIR doesn't work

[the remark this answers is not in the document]
duembgen: (so basically replace "any" by "all" in code snippet above)

PR #167, opened 2020-05-12
Vectorization of MUSIC 

Womac:

- [ x] Have you run the tests by doing `nosetests` or `py.test` at the root of the repo ?
- [x ] Last but not least, did you document the proposed change in the CHANGELOG file ? It should go under "Unreleased".

Happy PR :smiley:

I added vectorisation in the hidden functions of MUSIC to speed it up - I compared the obtained results. The maximum difference in the obtained grid values in the jupyter notebook example was 
3.885780586188048e-16 
Time Differences (measured with time):
MUSIC (mine)  0.020423173904418945
MUSIC (pyroomacoustics) 0.4093196392059326
I never contributed to a public repo yet, so please tell me if I missed something. 
Best, 
Wolfgang

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,14 @@ adheres to `Semantic Versioning <http://semver.org/spec/v2.0.0.html>`_.
 
 `Unreleased`_
 -------------
+Changed
+~~~~~~~
+
+- Added vectorised functions in MUSIC 
+- Use the vectorised functions in _process of MUSIC
+
+`0.3.1`_ - 2020-05-12
+---------------------
 
 Bugfix
 ~~~~~~
--- a/pyroomacoustics/doa/music.py
+++ b/pyroomacoustics/doa/music.py
@@ -7,7 +7,7 @@ from .doa import DOA
 
 class MUSIC(DOA):
     """
-    Class to apply MUltiple SIgnal Classication (MUSIC) direction-of-arrival 
+    Class to apply MUltiple SIgnal Classication (MUSIC) direction-of-arrival
     (DoA) for a particular microphone array.
 
     .. note:: Run locate_source() to apply the MUSIC algorithm.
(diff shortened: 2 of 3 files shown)


fakufaku:
Ouch! I left you hanging with a great PR! This can definitely be merged I think, it looks great! Thank you so much! Impressive speed up!

There is just a little thing. You are using the `@` operator for multiplication and it breaks python 2.7. I know 2.7 is sunsetting, but so as to not break things for people who might still be using faster than necessary, would you be ok with using `matmul` instead ?

Womac:
Thank you for the feedback. I am happy to contribute.

I changed the code such that it is compatible to python2.7 
The CI, however, gives the error 

'''An unexpected error has occurred. Conda has prepared the above report.'''

I think this error is not caused by me.

fakufaku:
Thanks! And yes, I should have mentioned the build on mac os with py27 seems broken...

Womac:
No worries. 
I have two more issues. I took mostly care of the vectorization of the code and implemented it accordingly to the old non-vectorised version. 
1. TOPS and CSSM still use the old non-vectorised versions. I did not have time yet to change these such that they use the new version [of the hidden functions in MUSIC, should not be too much work]. Afterwards the non-vectorised versions could be removed.  
2. In the code ` cross = identity - np.matmul(Es,np.moveaxis(np.conjugate(Es),-1,-2)) `[based on signal space eigenvectors Es] and in the paper   `cross = np.matmul(En,np.moveaxis(np.conjugate(En),-1,-2)) `[based on noise space Eigenvectors] are computed differently. I tried both methods, both work [in the test notebook]. Is there some advantage in the coded method? Otherwise, to avoid confusions, we maybe should stick with the method described in the paper?  Paper: [https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=1143830](url) Equation (6).

PR #168, opened 2020-05-13
Vectorization of SRP

flexo12:

Happy PR :smiley:

I have vectorised the matrix operations of SRP to speed them up. 
Best, Felix

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Changed
 - added vectorised matrix operations in SRP
 
 `0.3.3`_ - 2020-05-13
+---------------------
 
 Bugfix
 ~~~~~~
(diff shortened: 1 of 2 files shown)


fakufaku:
Hi, thank you so much for a great PR! Speed up is much appreciated for the DOA algorithms! This can definitely be merged I think!

There is just a small point. Would you mind replacing the `@` for matrix multiplication by `matmul` throughout ? This would allow to keep thing compatible for the (hopefully few) people who are still using python 2.7. Sorry for the annoyance!

fakufaku:
Hi @flexo12 , before trying to merge your code I was just running some tests and noticed that the vectorized version runs approximately twice slower than the original code. Have you tried to run some benchmark to assess the performance ?

fakufaku:
Something to take into account is that `self.mode_vec` is not a true numpy array, it is a wrapper that I wrote to do lazy computation of the mode vectors. If not, the precomputed array would be extremely large when using a lot of grid points. This might be what is slowing down the vectorized version (although I haven't checked). I would recommend using the [line_profiler](https://github.com/pyutils/line_profiler) to find out which part of the code is slow.

fakufaku:
I'll close this since there has been no updates for a while.

PR #175, opened 2020-07-02
[WIG] Support source directivities

fakufaku:
**Work in Progress**

This branch contains work towards support source directivities in pyroomacoustics.

TODO
-------

[ ] Python API to provide directivities to sources
[ ] Compute source emission angle in ISM shoebox
[ ] Compute source emission angle in ISM polyhedra
[X] Spherical distributions: uniform `pyroomacoustics.random.uniform_spherical`
[X] Spherical distributions: power `pyroomacoustics.random.power_spherical`
[ ] Spherical distributions: mixtures
[X] Usual patterns (figure-of-eight, cardioid, etc)
[ ] File based (e.g. CLF: common loudspeaker format)
[ ] Match power between ISM and ray tracing

Final checks
-------------

Happy PR :smiley:

### DIFF
--- a/pyroomacoustics/directivities.py
+++ b/pyroomacoustics/directivities.py
@@ -1,0 +1,183 @@
+import numpy as np
+from pyroomacoustics.utilities import Options
+
+
+class DirectivityPattern(Options):
+    FIGURE_EIGHT = "figure_eight"
+    HYPERCARDIOID = "hypercardioid"
+    CARDIOID = "cardioid"
+    SUBCARDIOID = "subcardioid"
+    OMNI = "omni"
+
+
+PATTERN_TO_CONSTANT = {
+    DirectivityPattern.FIGURE_EIGHT: 0,
+    DirectivityPattern.HYPERCARDIOID: 0.25,
+    DirectivityPattern.CARDIOID: 0.5,
+    DirectivityPattern.SUBCARDIOID: 0.75,
+    DirectivityPattern.OMNI: 1.0
+}
+
+
+class DirectionVector(object):
+    """
+    Parameters
+    ----------
+    azimuth : float
+    colatitude : float, optional
+        Default to PI / 2, only XY plane.
+    degrees : bool
+        Whether provided values are in degrees (True) or radians (False).
+    """
+    def __init__(self, azimuth, colatitude=None, degrees=True):
+        if degrees is True:
+            azimuth = np.radians(azimuth)
+            if colatitude is not None:
+                colatitude = np.radians(colatitude)
+        self._azimuth = azimuth
+        if colatitude is None:
+            colatitude = np.pi / 2
+        assert colatitude <= np.pi and colatitude >= 0
+        self._colatitude = colatitude
+
+    def get_azimuth(self, degrees=False):
+        if degrees:
+            return np.degrees(self._azimuth)
+        else:
+            return self._azimuth
+
+    def get_colatitude(self, degrees=False):
+        if degrees:
+            return np.degrees(self._colatitude)
+        else:
+            return self._colatitude
+
+
+class Directivity(object):
+    """
+
+    Parameters
+    ----------
+    orientation : DirectionVector
+        Indicates direction of the pattern.
+    pattern : str
+        One of directivities in `DirectivityPattern`.
+        TODO : support arbitrary directivities.
+
+    """
+    def __init__(self, orientation, pattern):
+        assert pattern in DirectivityPattern.values()
+        assert isinstance(orientation, DirectionVector)
+        self._orientation = orientation
+        self._pattern = pattern
+        self._p = PATTERN_TO_CONSTANT[pattern]
+
+    def get_directivity_pattern(self):
+        return self._pattern
+
+    def get_azimuth(self, degrees=True):
+        return self._orientation.get_azimuth(degrees)
+
+    def get_colatitude(self, degrees=True):
+        return self._orientation.get_colatitude(degrees)
+
+    def get_response(self, direction):
+        """
+        Get response for a single direction.
+
+        TODO : vectorize
+
+        Parameters
+        ----------
+        direction : DirectionVector
+            Direction for which to compute gain
+
+        """
+        assert isinstance(direction, DirectionVector)
+        if self._pattern == DirectivityPattern.OMNI:
+            return 1
+        else:
+            # inspiration from Habets: https://github.com/ehabets/RIR-Generator/blob/5eb70f066b74ff18c2be61c97e8e666f8492c149/rir_generator.cpp#L111
+            # we use colatitude instead of elevation
+            gain = np.sin(self._orientation.get_colatitude()) * \
+                   np.sin(direction.get_colatitude()) * \
+                   np.cos(
+                       self._orientation.get_azimuth() -
+                       direction.get_azimuth()
+                   ) + \
+                   np.cos(self._orientation.get_colatitude()) * \
+                   np.cos(direction.get_colatitude())
+            return np.abs(self._p + (1 - self._p) * gain)
+
+    def plot_response(self, azimuth, colatitude=None, degrees=True):
+        """
+        Parameters
+        ----------
+        azimuth : array_like
+            Azimuth values for plotting.
+        colatitude : array_like, optional
+            Colatitude values for plotting. If not provided, 2D plot.
+        degrees : bool
+            Whether provided values are in degrees (True) or radians (False).
+        """
+
+        try:
+            import matplotlib.pyplot as plt
+        except ImportError:
+            import warnings
+            warnings.warn('Matplotlib is required for plotting')
+            return
+
+        fig = plt.figure()
+        if colatitude is not None:
+
+            # compute response
+            gains = np.zeros(shape=(len(azimuth), len(colatitude)))
+            for i, a in enumerate(azimuth):
+                for j, c in enumerate(colatitude):
+                    direction = DirectionVector(
+                        azimuth=a,
+                        colatitude=c,
+                        degrees=degrees
+                    )
+                    gains[i, j] = self.get_response(direction)
+
+            # create surface plot, need cartesian coordinates
+            if degrees:
+                azimuth = np.radians(azimuth)
+                colatitude = np.radians(colatitude)
+            AZI, COL = np.meshgrid(azimuth, colatitude)
+
+            X = gains.T * np.sin(COL) * np.cos(AZI)
+            Y = gains.T * np.sin(COL) * np.sin(AZI)
+            Z = gains.T * np.cos(COL)
+
+            ax = fig.add_subplot(1, 1, 1, projection='3d')
+            ax.plot_surface(X, Y, Z)
+            ax.set_title("{}, azimuth={}, colatitude={}".format(
+                self.get_directivity_pattern(),
+                self.get_azimuth(),
+                self.get_colatitude()
+            ))
+            ax.set_xlabel("x")
+            ax.set_ylabel("y")
+            ax.set_zlabel("z")
+            ax.set_xlim([-1, 1])
+            ax.set_ylim([-1, 1])
+            ax.set_zlim([-1, 1])
+
+        else:
+
+            # compute response
+            gains = np.zeros(len(azimuth))
+            for i, a in enumerate(azimuth):
+                direction = DirectionVector(azimuth=a, degrees=degrees)
+                gains[i] = self.get_response(direction)
+
+            # plot
+            ax = plt.subplot(111, projection="polar")
+            if degrees:
+                angles = np.radians(azimuth)
+            else:
+                angles = azimuth
+            ax.plot(angles, gains)
--- a/pyroomacoustics/directivities.py
+++ b/pyroomacoustics/directivities.py
@@ -1,26 +1,18 @@
 import abc
 import numpy as np
+from enum import Enum
 
 from pyroomacoustics.doa import spher2cart
-from pyroomacoustics.utilities import Options, requires_matplotlib, \
-    all_combinations
+from pyroomacoustics.utilities import requires_matplotlib, all_combinations
 
 
-class DirectivityPattern(Options):
-    FIGURE_EIGHT = "figure_eight"
-    HYPERCARDIOID = "hypercardioid"
-    CARDIOID = "cardioid"
-    SUBCARDIOID = "subcardioid"
-    OMNI = "omni"
-
-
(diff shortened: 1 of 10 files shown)


ebezzam:
Idea for API to provide directivities to sources:

```
# after creating a room
room.add_source(
    position=position,
    directivity=directivity,      # could be str for one of fixed patterns or an arbitrary pattern provided through a matrix
    orientation=[azimuth, elevation]
)
```

And for the `SoundSource` object, adding arguments for `directivity` and `orientation` in the constructor and methods to set them.

fakufaku:
I think that is one idea for the API in the Room object. But I was thinking more of an abstract class to offer a uniform interface for both Sources and Microphones. I think the constructor arguments might be different depending on the type of the source/microphone. In the beginning elevation/azimuth are enough, but later we might want to provide measurements from head-related transfer function, so we need to provide the path to a file, for example.

```python
import abc
class DirectiveObject(abc.ABC):

    @abstractmethod
    get_response(self, dir_vector):
        """
        Parameters
        ------------
        dir_vector: array_like, shape (n_dim,) or (n_dim, n_vectors)
            Unit vectors indicating the direction where the response is required

        Returns
        --------
        response: array_like, shape (n_freq,) or (n_freq, n_vectors)
            The response of the source or microphone at the required directions.
            Note: It is not completely clear to me yet what exactly we expect. Is it possible that these
                are impulse responses in some cases ? I believe yes. We should take this into account.
        """
        pass
```

Then we can define other types of objects

```python
class Cardioid:
    """ An object implementing a Cardioid directive response """
    def __init__(self, azimuth=0., colatitude=None, gain=2.0):  # <- do we stick to colatitude as in DOA ?
        if not elevation:
            self.colatitude = np.pi / 2
        else:
            self.colatitude = colatitude
        self.azimuth = azimuth
        self.gain

        self.unit_v = np.array([
            np.cos(self.azimuth) * np.sin(self.colatitude),
            np.sin(self.azimuth) * np.sin(self.colatitude),
            np.cos(self.colatitude),
        ])

    def __call__(self, dir_vector):
        return self.gain * (1.0 + np.matmul(self.unit_v[None, :], dir_vector))

class SoundSource(DirectiveObject):
    def __init__(self, position, directivity=None, *args, **kwargs):
        self.position = position
        if not directivity:
            self.directivity = Omnidirectional()
        else:
            self.directivity = directivity

    def get_response(self, dir_vector):
        return self.directivity(dir_vector)
```

So that we could use it in the following way

```python
room.add(SoundSource(position=pos, directivity=Cardioid(azimuth=np.pi / 4, colatitude=np.pi / 3))
```
or 
```python
room.add_source(
    position=position,
    directivity=Cardioid(azimuth=np.pi / 4, colatitude=np.pi / 3),
)
``` 
An alternative I first thought about would be to have derived classes like
```python
class CardioidObject(DirectiveObject):
    ...

class CardioidSource(CardioidObject):
    ...
```
but I thought this might require too much boilerplate as the constructor of `CardioidSource` and friends would need to be rewritten everytime with both arguments for `SoundSource` and the directivity...

ebezzam:
@fakufaku I've added something similar to the `DirectiveObject` that you proposed, only for closed form directivity patterns at the moment. And could still use some vectorization, just wanted to make sure I got the equation down right and some scripts to visualize. 

I think keeping `colatitude` is good for consistency. I've added this notion of a `DirectionVector`. Let me know what you think.

And here are some directivity plots from some scripts I added in `examples` 🙂 

comparing_patterns
3d

pyroomacoustics/directivities.py
```diff
@@ -0,0 +1,183 @@
...
+
+class DirectivityPattern(Options):
+    FIGURE_EIGHT = "figure_eight"
+    HYPERCARDIOID = "hypercardioid"
+    CARDIOID = "cardioid"
+    SUBCARDIOID = "subcardioid"
+    OMNI = "omni"
+
+
+PATTERN_TO_CONSTANT = {
```
fakufaku, line 13: If this is a purely local constant, should we prefix it with `_` ?

pyroomacoustics/utilities.py
```diff
@@ -738,6 +738,16 @@ def goertzel(x, k):
     return y
 
 
+class Options(object):
```
fakufaku, line 741: I was just wondering if something equivalent exists in [collections](https://docs.python.org/3.8/library/collections.html). Also, maybe we could add a docstring to make the purpose of this class more obvious.

pyroomacoustics/directivities.py
```diff
@@ -0,0 +1,183 @@
...
+        ----------
+        azimuth : array_like
+            Azimuth values for plotting.
+        colatitude : array_like, optional
+            Colatitude values for plotting. If not provided, 2D plot.
+        degrees : bool
+            Whether provided values are in degrees (True) or radians (False).
+        """
+
+        try:
```
fakufaku, line 124: Just a thought for the future, but the check for existence of matplotlib could be implemented as a decorator.

```python

@requires_matplotlib
def my_plot_func(*args, **kwargs):
    plt.plot(...)
```

pyroomacoustics/directivities.py
```diff
@@ -0,0 +1,183 @@
...
+    pattern : str
+        One of directivities in `DirectivityPattern`.
+        TODO : support arbitrary directivities.
+
+    """
+    def __init__(self, orientation, pattern):
+        assert pattern in DirectivityPattern.values()
+        assert isinstance(orientation, DirectionVector)
+        self._orientation = orientation
+        self._pattern = pattern
```
fakufaku, line 72: Here I would change the name to `_pattern_str` or `_pattern_name` to make it obvious that this is the name of the type rather than the actual curve of the pattern.

pyroomacoustics/directivities.py
```diff
@@ -0,0 +1,183 @@
...
+        self._p = PATTERN_TO_CONSTANT[pattern]
+
+    def get_directivity_pattern(self):
```
fakufaku, line 75: Here we could use the `__repr__` or `__str__` special methods.

Also, recently I quite like the property decorator for getter and setter. Here it would be like this:
```python
@property
def directivity_pattern(self):
    return self._pattern
```

pyroomacoustics/directivities.py
```diff
@@ -0,0 +1,183 @@
...
+            return self._azimuth
+
+    def get_colatitude(self, degrees=False):
+        if degrees:
+            return np.degrees(self._colatitude)
+        else:
+            return self._colatitude
+
+
+class Directivity(object):
```
fakufaku, line 56: I think `Directivity` is a little too general for this class since it refers only to the Cardioid/Hypercardioid, etc family of patterns. Does this family have a name ?

Also, I think it would be nice to create derived classes that we directly call for the different types.
```python
class Cardioid(Directivity):
    def __init__(self, orientation=None, ...)
        super().__init__(orientation=orientation, pattern=DirectivityPattern.CARDIOID)
```

pyroomacoustics/directivities.py
```diff
@@ -0,0 +1,183 @@
...
+PATTERN_TO_CONSTANT = {
+    DirectivityPattern.FIGURE_EIGHT: 0,
+    DirectivityPattern.HYPERCARDIOID: 0.25,
+    DirectivityPattern.CARDIOID: 0.5,
+    DirectivityPattern.SUBCARDIOID: 0.75,
+    DirectivityPattern.OMNI: 1.0
+}
+
+
+class DirectionVector(object):
```
fakufaku, line 22: For this one, I am a little bit ambivalent. It makes things quite clear and let us run some checks, but at the same time, it makes the machinery more heavy, compared to just using numpy arrays.

I think there are different solutions:
1. Just use a function to create unit vectors from spherical coordinates. I think we already have them in the doa sub-package, so we could just pull them somewhere more obvious.
2. Make `DirectionVector` a [sub-class of ndarray](https://numpy.org/doc/stable/user/basics.subclassing.html). This is quite neat. We can make sure the vector is normalized in the constructor, and then, we can still use this class as input to any numpy method. We can also add methods to convert back to azimuth/colatitude. This could also allow us to deal with the vectorized case by for example allowing extra dimensions.
3. Use plain numpy arrays...

fakufaku:
Also throwing this here so that we don't forget: for cardioid and friends, is the response for the energy or the amplitude ? i.e., do we need to take a square root before multiplying the RIR ?

fakufaku:
@ebezzam Turns out there is a reference for source directivies in the shoebox model: [link](https://www2.ak.tu-berlin.de/~akgroup/ak_pub/2018/000458.pdf). Haven't read in detail yet, but as far as I can tell, this is exactly what we had in mind.

pyroomacoustics/directivities.py
```diff
@@ -0,0 +1,183 @@
...
+            Direction for which to compute gain
+
+        """
+        assert isinstance(direction, DirectionVector)
+        if self._pattern == DirectivityPattern.OMNI:
+            return 1
+        else:
+            # inspiration from Habets: https://github.com/ehabets/RIR-Generator/blob/5eb70f066b74ff18c2be61c97e8e666f8492c149/rir_generator.cpp#L111
+            # we use colatitude instead of elevation
+            gain = np.sin(self._orientation.get_colatitude()) * \
```
fakufaku, line 102: This may be more efficient in the vector form. Instead of using azimuth/colatitude, we store the unit vector pointing  in the direction and then we only need one inner product. This way, cos/sin only need to be called once in the constructor.

pyroomacoustics/directivities.py
```diff
@@ -0,0 +1,183 @@
...
+                   np.sin(direction.get_colatitude()) * \
+                   np.cos(
+                       self._orientation.get_azimuth() -
+                       direction.get_azimuth()
+                   ) + \
+                   np.cos(self._orientation.get_colatitude()) * \
+                   np.cos(direction.get_colatitude())
+            return np.abs(self._p + (1 - self._p) * gain)
```
fakufaku, line 110: Also, it is unclear if we need the abs here. In actual microphones, there will be a phase shift. Maybe an option ? Or do we delegate responsibility for the sign to the caller of the function ?

pyroomacoustics/directivities.py
fakufaku: Maybe the class can be called `CardioidFamily`.

fakufaku: And, can we make it also a simple function, that may come in handy for what I am currently doing with the ray tracing.
```python
def cardioid_func(x, loc, coef, gain):
     """
     Parameters
     x: array_like, shape (..., n_dim)
          The input
     loc: array_like, shape (n_dim)
          The direction vector (should be normalized, but no need to check in the function)
     coeff: float
          The parameter of the cardioid
     gain: float
          The gain
     """
     return gain * (coeff + (1 - coeff) * np.matmul(x, loc)
```
And we will also need to have functions for the volume (3D) and area (2D) of cardioids for the energy matching.

pyroomacoustics/utilities.py
ebezzam: I couldn't see anything in collections that would have equivalent functionality, do you know of something? I've added some docstring in any case.

pyroomacoustics/directivities.py
ebezzam: added one!

ebezzam:
> Also throwing this here so that we don't forget: for cardioid and friends, is the response for the energy or the amplitude ? i.e., do we need to take a square root before multiplying the RIR ?

If I refer to [Habet's code](https://github.com/ehabets/RIR-Generator/blob/5eb70f066b74ff18c2be61c97e8e666f8492c149/rir_generator.cpp#L423), it seems like this is the response for the amplitude, as the gain is used without square root when multiplying with the RIR.

pyroomacoustics/utilities.py
fakufaku: Right, maybe we can achieve the same thing by using an [enum](https://docs.python.org/3/library/enum.html). Then, we don't provide a string as argument, but something like "DirectivityPattern.CARDIOID".

pyroomacoustics/directivities.py
ebezzam: I've added an option `magnitude` for returning the absolute value.

pyroomacoustics/directivities.py
ebezzam: I've switched to computing the response with Cartesian coordinates so now it is vectorized.

pyroomacoustics/directivities.py
ebezzam: done in most recent commits!

pyroomacoustics/utilities.py
ebezzam: Enum looks good 👍 I've switched to this.

pyroomacoustics/directivities.py
ebezzam: I agree, it makes it clunky and harder to vectorize. I've kept this `DirectionVector` for when creating a `CardioidFamily` object so that `azimuth` and `colatitude` (and the corresponding unit vector) are bundled into a single object.

ebezzam:
@fakufaku I've taken into account a lot of your suggestions. Some things still to do for directivities / cardioids:
- Functions for area and volume of cardioids.
- Proper unit tests, at the moment I'm using the scripts that I've added in the `examples` folder to test. But it would be nicer to have formal unit tests.

Let me know if you have any other suggestions for the directivity API. But if it looks good then I can move forward to using this for ISM 🙂

pyroomacoustics/directivities.py
fakufaku: Cool! I am now using this! Small detail, the shape of `x` in the docstring doesn't match the implementation. The docstring should say `(n_dim, n_points)` or something like that.

pyroomacoustics/directivities.py
fakufaku: Sounds good!

fakufaku:
Looks good to me! I think we can move forward to integrating this in the simulator (ISM for now). As we do this, it would be nice to already think about the future and add code to handle non-shoebox/ray tracing later. What I mean is preparing the code path, but raise `NotImplementedError` for the stuff that is not there yet.

For the tests, let's do that as we go. The most important is to make sure the code is called during the tests, so maybe create a simple test file that creates these objects and calls their methods. Where possible/easy, test for correctness, etc.

fakufaku:
@ebezzam Could you confirm that this PR is obsolete and that I can close it ?

fakufaku:
Ping @ebezzam again. This looks like an obsolete PR now that Satvik's project has been merged. I'll close it for now.

PR #180, opened 2020-07-14
[WIP] Source directivity for ray tracing

fakufaku:
I have started work for the source directivity of the ray tracer. The idea is to implement the drectivity as a random sampling of the rays on the sphere from the distribution induced by the source directivity. This requires to sync a number of features with the source directivity for the ISM, in particular, the energy should be matched, which requires to know the integral of the source directivity patterns. Ideally, we should store all of this in easy use classes.

[ ] Energy matching with ISM
[ ] Add interface to provide ray energies along with the directions to the C++ engine.
[ ] Tests: How do we test the distributions ? Use statistical tests ? How do we compute averages on the sphere ?
[ ] docstrings
[ ] Doc building ?
[ ] Add to CHANGELOG

### DIFF
--- a/pyroomacoustics/random/sampler.py
+++ b/pyroomacoustics/random/sampler.py
@@ -1,0 +1,81 @@
+import numpy as np
+from numpy.random import default_rng
+
+from . import distributions
+from .spherical import uniform_spherical
+
+
+class RejectionSampler:
+    def __init__(self, desired_func, proposal_dist, scale=None):
+
+        if not scale:
+            self._scale = 1.0
+        else:
+            self._scale = scale
+
+        self._desired_func = desired_func
+        self._proposal_dist = proposal_dist
+
+        self._dim = self._proposal_dist.dim
+
+        # get a random number generator
+        self._rng = default_rng()
+
+    @property
+    def dim(self):
+        return self._dim
+
+    def __call__(self, size=None):
+
+        if not size:
+            size = 1
+
+        flat_size = int(np.prod(size))  # flat size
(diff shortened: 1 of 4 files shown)


pyroomacoustics/random/sampler.py
```diff
@@ -0,0 +1,104 @@
...
+    loc: array_like
+        The unit vector pointing in the main direction of the cardioid
+    """
+
+    def __init__(self, loc=None, coeff=0.5):
+        super().__init__(loc=loc)
+        self._coeff = coeff
+
+    def _pattern(self, x):
+        return np.abs(self._coeff + self._coeff * np.matmul(x, self._loc))
```
ebezzam, line 104: It shouldn't be `np.abs(self._coeff + (1 - self._coeff) * np.matmul(x, self._loc))`? And as you said in the other PR, I don't think we need the `abs`.

In any case, I've added the ["one-shot" function](https://github.com/LCAV/pyroomacoustics/blob/7a6486c8b2f3402b0529180d8f6a2dd169f6b8dd/pyroomacoustics/directivities.py#L181) that you asked for, which you can call like so:

```python
from pyroomacoustics.directivities import cardioid_func

...

cardioid_func(x, self._loc, self._coeff)
```

fakufaku: You are right about the `(1 - coeff)`, but I have now replaced this by a call to your function 👍 

Here we need the `abs` because we are drawing samples from a distribution that needs to be non-negative.

PR #183, opened 2020-08-06
Normalization of Fractional Delay Filter

ntcnjdbxz:
Current implementation of the Room Impulse Responses (RIRs) contains a critical bug. Namely, there is no normalization of the fractional delay filter. A delay filter has to add a delay without attenuation of the signal. Hence, it gives unexpected amplitudes of the RIRs. It is known that the signal strength decays over distance, without normalization this can not be achieved. On the attached figures you can see results of the current implementation and after improvement.
Simulated_setup
Simulated_RIRs_before
Simulated_RIRs_after

### DIFF
--- a/pyroomacoustics/build_rir.pyx
+++ b/pyroomacoustics/build_rir.pyx
@@ -64,7 +64,7 @@ def fast_rir_builder(
     cdef int lut_pos, i, f, time_ip
     cdef float x_off, x_off_frac, sample_frac
 
-    for i in range(n_times):
+       for i in range(n_times):
         if visibility[i] == 1:
             # decompose integer and fractional delay
             sample_frac = fs * time[i]
@@ -77,8 +77,21 @@ def fast_rir_builder(
             x_off = (x_off_frac - lut_gran_off)
             lut_pos = lut_gran_off
             k = 0
-            for f in range(-fdl2, fdl2+1):
-                rir[time_ip + f] += alpha[i] * hann[k] * (sinc_lut[lut_pos] 
+            for k in range(2*fdl2+1):
+                rir_tmp[k] = hann[k] * (sinc_lut[lut_pos]
                         + x_off * (sinc_lut[lut_pos+1] - sinc_lut[lut_pos]))
                 lut_pos += lut_gran
-                k += 1
+
+
+            max_tmp = 0
+            for f in range(2*fdl2+1):
+                if rir_tmp[f] > max_tmp:
+                    max_tmp = rir_tmp[f]
+
+            for f in range(2*fdl2+1):
+                rir_tmp[f] = alpha[i]*rir_tmp[f]/max_tmp
+
+            k = 0
+            for f in range(-fdl2, fdl2+1):
+                rir[time_ip + f] += rir_tmp[k]
+                k+=1
--- a/pyroomacoustics/build_rir.pyx
+++ b/pyroomacoustics/build_rir.pyx
@@ -63,6 +63,8 @@ def fast_rir_builder(
     cdef double [:] hann = np.hanning(fdl)
     cdef int lut_pos, i, f, time_ip
     cdef float x_off, x_off_frac, sample_frac
+    cdef double max_tmp
+    cdef double [:] rir_tmp = np.zeros(2*fdl2+1)
 
     for i in range(n_times):
         if visibility[i] == 1:
--- a/pyroomacoustics/build_rir.pyx
+++ b/pyroomacoustics/build_rir.pyx
@@ -64,7 +64,7 @@ def fast_rir_builder(
     cdef int lut_pos, i, f, time_ip
     cdef float x_off, x_off_frac, sample_frac
 
-       for i in range(n_times):
+    for i in range(n_times):
         if visibility[i] == 1:
             # decompose integer and fractional delay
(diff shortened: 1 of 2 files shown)


fakufaku:
Hi @ntcnjdbxz , thanks for the PR. However, I don't believe there is a bug in the code. Please bear with me for a moment. Here is what happens in the processing chain.

1) The delays at the microphone are computed according to the distance, which is a continuous value
2) Before sampling, the RIR is just a train of Dirac
3) To discretize the RIR, we apply a low pass filter. Note that this low-pass filter is located at microphone and is independent of the distance to the source. In pyroomacoustics, we use a windowed sinc as a low-pass filter
4) After convolving the continuous train of dirac with the windowed sinc, we uniformly sample the signal

Now, this is at this step that something counter-intuitive may happen. Imagine the two following cases
a) the delay at the microphone falls exactly on a sample
b) the delay falls exactly at the middle between two neighboring samples

In a), the amplitude at the peak will be exactly one, and the neighboring samples will all fall on zeros of the sinc function
In b), the two samples directly on the side of the Dirac will have the same value of sinc(pi / 2) / (pi / 2), and the other samples will also have a non-zero value

Now, imagine that two dirac arrive one after another, however, the first dirac to arrive (closer to the microphone) follow pattern b), while the second (further from the mic) follows pattern a). If the distance are not too different, the difference in amplitude due to the difference in sampling (a or b), may be larger than the difference due to distance. In that case, the largest sample due to the first Dirac will be smaller than the one due to the second.

While this may be surprising, recall that the dirac has been spread due to the low-pass filter. If you consider that when the dirac falls between samples there are larger side-lobes (the sinc decay slowly), then the (averaged) energy due to the first delay is in fact larger than for the second, even though its maximum sample appears smaller.

I hope this clarifies the situation a little bit.

ntcnjdbxz:
Nevertheless, parameter "alpha" in "soundsource.py" defines attenuation of the signal according to distance. Delay filter is designed based ONLY on delay in samples and without connection to the calculated attenuation for the current distance. Which means, when delay filter produces extra attenuation, it spoils calculated parameter "alpha".

ntcnjdbxz:
Moreover, according to your explanation the attenuation of the delay filter differs (randomly!). Which gives even bigger difference between actual attenuation "alpha" for 1-st microphone and 2-nd and so on. And now imagine that you normalized delay filter (no attenuation at all). Hence, this described by you ambiguity, disappear.  Thus, we get normal amplitude fade over distance. Of course if you don't want to model random obstacles on the way of sound wave.

fakufaku:
> Moreover, according to your explanation the attenuation of the delay filter differs (randomly!). Which gives even bigger difference between actual attenuation "alpha" for 1-st microphone and 2-nd and so on.

This is in fact quite the opposite. Fractional delay filters are a type of [all pass filter](https://en.wikipedia.org/wiki/All-pass_filter), that is they do not change the amplitude of the frequency content of the signal. Fractional delays are in addition linear phase, meaning they change the delay of each frequency by the same amount, i.e., they are just a delay of the signal. 

In fact, by normalizing by the largest component, as you suggest to do you introduce a gain that depends on the delay. This can be observed by plotting the frequency response of the filters with and without the modification. Note how on the right the gain of the filter depends heavily on the delay.
```python
import numpy as np

import pyroomacoustics as pra


def get_response(f, L, n_fft):
    fft_buffer = np.zeros(n_fft)
    fft_buffer[: L // 2 + 1] = f[L // 2 :]
    fft_buffer[-L // 2 + 1 :] = f[: L // 2]

    response = np.fft.rfft(fft_buffer)

    return np.abs(response) ** 2


if __name__ == "__main__":

    delays = [0.0, 0.13, 0.32, 0.39, 0.5]
    L = pra.constants.get("frac_delay_length")
    n_fft = 512

    fig, (ax1, ax2) = plt.subplots(1, 2)

    fig.suptitle("Frequency Response of Fractional Delay Filters")
    ax1.set_title("In pyroomacoustics")
    ax2.set_title("With division by max")

    for d in delays:

        f = pra.fractional_delay(d)
        f_m = f / np.max(f)

        ax1.plot(get_response(f, L, n_fft))
        ax2.plot(get_response(f_m, L, n_fft))

    for ax in [ax1, ax2]:
        ax.set_xlabel("frequency")

        ax.set_xticks([0, n_fft // 2])
        ax.set_xticklabels([0, "fs/2"])
        ax.set_ylim([0, 2.6])

        ax.legend([f"{d:.2f}" for d in delays], title="delay", loc="lower left")

    ax1.set_ylabel("|FFT(x)|$^2$")
    ax2.set_yticks([])

    fig.tight_layout(rect=[0, 0.00, 1, 0.95])

    fig.savefig("frac_delay_response.png", dpi=150)
    plt.show()
```
frac_delay_response.

fakufaku:
Just to add a little bit of insight, you need to imagine that we are filtering the full continuous impulse response with a single, continuous low-pass filter before sampling. The fractional delay filter is just an artefact of the implementation. As a consequence, it makes sense that the frequency response (and gain) should be uniform across the impulse response. This is explained to some extent in the [paper](https://arxiv.org/abs/1710.04196) (equations (1) and (2)). You can derive it by computing the convolution integral of the dirac pulse train of the continuous impulse response with the continuous filter of eq. (2).

PR #194, opened 2020-10-27
Improve the computation of the auxiliary variables in AuxIVA and ILRMA

mori97:
I improved the computation of the auxiliary variables of AuxIVA and ILRMA.
The original implementation
```python
V[:, :, :] = np.matmul(
    (X * r_inv[None, s, None, :]), np.conj(X.swapaxes(1, 2)) / n_frames
)
```
performs division operation `n_freq` x `n_frames` x `n_chan` times.
I modified this computation as
```python
V[:, :, :] = np.matmul(
    (X * r_inv[None, s, None, :]), np.conj(X.swapaxes(1, 2))
) / n_frames
```
which only performs division operation  `n_freq` x `n_chan` x `n_chan` times.
Since the runtime of AuxIVA is dominated by the computation of `V`,
this modification makes AuxIVA run much faster.
I tested AuxIVA by using `%timeit` of iPython with `n_frames = 300`, `n_freq = 2049`, `n_chan = 3`, `n_src = 3`.
Before:
```
4.7 s ± 2.65 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
After:
```
4.04 s ± 4.74 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
The implementation of ILRMA is also modified in the same way.

(This PR did not modify any documents.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,12 @@ Bugfix
 
 - Fixes typo in a docstring
 
+Changed
+~~~~~~~
+
+- Improve the computation of the auxiliary variables in AuxIVA and ILRMA.
+  Unnecessary division operations are reduced.
+
 `0.4.2`_ - 2020-09-24
 ---------------------
 
--- a/pyroomacoustics/bss/auxiva.py
+++ b/pyroomacoustics/bss/auxiva.py
@@ -225,8 +225,8 @@ def auxiva(
             # Compute Auxiliary Variable
             # shape: (n_freq, n_chan, n_chan)
             V[:, :, :] = np.matmul(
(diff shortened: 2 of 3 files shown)


jazcarretao:
Good spot and thanks for the profiling! Have you run the tests just in case?

mori97:
I ran the tests with `nosetests`.
Also, I have confirmed the outputs of `auxiva` and `ilrma` do not change after this modification by `numpy.all_close`.

jazcarretao:
Thanks. Looks good to me then.

fakufaku:
@mori97 Thank you so much for the contribution and sorry for the delay in merging!!! 👍 @jazcarretao Thanks for reviewing!

PR #197, opened 2020-11-20
BUG in SRP 

Womac:

- [] Are there docstrings ? Do they follow the [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) style ?

Happy PR :smiley:

Hey, 

I think there is a BUG in the computation of SRP (also, when using several frequencies, the BUG should average out). When the final SRP cost is computed, the current code uses a `np.abs`.  As `srp_cost` is obtained, by summing the upper triangle per frequency of the PSD matrix, there will be imaginary parts left [which should average out when averaging over the frequencies]. However, for a single frequency and two microphones, the result would be ones in any direction. Consequently, I think it is necessary to replace the `np.abs` via `np.real`. 

If we would sum over the whole PSD matrix, the imaginary parts would cancel each other, as the matrix is hermetian.

Best, 
Wolfgang


fakufaku:
Hi @Womac , did this actually trip any of your code, or do you have an example where this fails as you describe above ?

My understanding is that the code is currently correct. The way the computation is done is a little bit confusing, so here is how the math goes.
Let `a` be the mode vector (size M) and `R` the covariance matrix (size MxM). Both are complex. Let `^T` and `^H` be the conjugate and hermitian conjugate operators, respectively, and `tr( . )` be the trace opeartor.
Then, the SRP cost is given by
```
a^H R a = tr(a a^H R) = 1^T ( (a a^H)^T * R ) 1
```
Note that `a^H R a` is a quadratic form, which means that it is always both real and positive. This is why the use of `abs` and `real` should be interchangeable.
But now that I look more closely into it, I am a little bit troubled by the `triu` use here, so I'll look into it.

fakufaku:
After a quick check, it seems indeed that the use of `triu` here may be problematic. I just checked, and without the `triu`, the imaginary part of `sum_val` is zero up to machine precision (~`1e-12`), i.e. it is real (positive). With `triu`, `sum_val` is a complex value. @ebezzam do you remember how we came up with use the upper triangular part only ?

fakufaku:
I could check that the current computation of the cost function leads to an error for two microphones and one frequency sub-band. Thanks @Womac ! It was however not sufficient to just replace `abs` by `real` to fix the computation, so I opened PR #199 to fix the problem.

fakufaku:
Moved to #199

Womac:
@fakufaku I think .real solves he problem. If the whole PSD is summed up, the imaginary parts of all elements should cancel out [as the matrix is hermetian]. The real parts will just double for the off-diagonal elements compared to when using triu. The diagonal elements are just a 'DC' offset as described in (http://www.glat.info/ma/av16.3/2000-DiBiaseThesis.pdf) 
I think the error comes from [page 80 in the PhD thesis] `Rlq(∆q −∆l)=Rql(∆l −∆q) ⇒ Rlq(∆q −∆l)+Rql(∆l −∆q)=2Rlq(∆q [−∆l)`,]
which has been applied in frequency domain in your code, without compensating the phase offset. Triu does not use the diagonal elements [which is fine], however, the 'SRP' for some directions then could be negative [which is fine], as the DC offset is not taken into account.

fakufaku:
@Womac I think you are indeed right and your proposed fix yields the correct cost. However, the implementation can be simplified a little bit to avoid computation of the imaginary part of the cost function which is always zero. This is what I have done in #199.

PR #199, opened 2020-11-24
Bug in computation of SRP cost function

fakufaku:
This PR corrects a bug in the computation of the cost function of the SRP-PHAT DOA algorithm.
The bug was originally reported in PR #197 .

This PR fixes the original bug, improves computational efficiency (although this was not really tested), and adds a test that catches the original bug.

Happy PR :smiley:

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -32,16 +32,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-    - name: Lint with flake8
-      run: |
-        # pip install flake8
-        # stop the build if there are Python syntax errors or undefined names
-        # flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-        # flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
-    #- name: Set PATH
-    #  run: echo "::add-path::C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\x64"
-    #  if: matrix.os == 'windows-latest'
     - name: Build package
       run: |
         python setup.py build_ext --inplace
--- a/pyroomacoustics/doa/srp.py
+++ b/pyroomacoustics/doa/srp.py
@@ -68,15 +68,11 @@ class SRP(DOA):
 
         self.num_pairs = self.M * (self.M - 1) / 2
 
-        # self.mode_vec = np.conjugate(self.mode_vec)
-
     def _process(self, X):
         """
         Perform SRP-PHAT for given frame in order to estimate steered response 
         spectrum.
         """
-
-        ones = np.ones(self.L.shape[1])
 
         srp_cost = np.zeros(self.grid.n_points)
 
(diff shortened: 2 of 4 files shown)


fakufaku:
@ebezzam Would you mind checking this PR please ?

pyroomacoustics/doa/srp.py
```diff
@@ -63,24 +87,34 @@ def _process(self, X):
...
         for k in self.freq_bins:
-            CC.append( np.dot(pX[:,k,:], np.conj(pX[:,k,:]).T) )
+            CC.append(np.dot(pX[:, k, :], np.conj(pX[:, k, :]).T))
         CC = np.array(CC)
 
+        M = self.L.shape[1]
+        ar = np.arange(M)
+        av = ar[:, None]
+        mask_diag = av == av.T
+        mask_triu = av <= av.T
```
ebezzam, line 97: this variable can be removed as it is not used

ebezzam: as well as `ones = np.ones(self.L.shape[1])` at the top of `_process`

pyroomacoustics/doa/srp.py
```diff
@@ -63,24 +87,34 @@ def _process(self, X):
...
             # compute the outer product along the microphone axis
-            mode_mat = np.conj(mode_vec[:,:,None]) * mode_vec[:,None,:]
-
-            # multiply covariance by mode vectors and sum over the frequencies
-            R = np.sum(CC * mode_mat, axis=0)
+            mode_mat = np.conj(mode_vec[:, :, None]) * mode_vec[:, None, :]
 
-            # Now sum over all distince microphone pairs
-            sum_val = np.inner(ones, np.dot(np.triu(R, 1), ones))
+            # method 2
```
ebezzam, line 111: this is very elegant! But maybe a more descriptive comment to explain the logic (it took me some time to parse out). Something like: "sum real part of upper triangle of CC * mode_mat as matrix is hermitian.

ebezzam: and do we need to add the diagonal components? as @Womac pointed out and on p. 80 of the thesis, the M autocorrelations are " independent of the steering delays". So from what I understand, this same offset should be present for all grid points. Which is why I think we used `np.triu(R, 1)` before, but the imaginary components were previously included, which was wrong.

ebezzam (changes requested):
Sorry for the delay, this was indeed a good bug catch! I've added some small comments.

.github/workflows/pythonpackage.yml
```diff
@@ -39,9 +39,9 @@ jobs:
         # flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         # flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
-    - name: Set PATH
-      run: echo "::add-path::C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\x64"
-      if: matrix.os == 'windows-latest'
+    #- name: Set PATH
```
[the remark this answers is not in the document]
fakufaku: Things have changed in the configuration of the windows runner of github actions. The commented line used to be necessary to correctly set the path on windows. I will delete this.

pyroomacoustics/doa/srp.py
```diff
@@ -36,15 +37,38 @@ class SRP(DOA):
...
+            mode=mode,
+            r=r,
+            azimuth=azimuth,
+            colatitude=colatitude,
+            **kwargs
+        )
+
+        self.num_pairs = self.M * (self.M - 1) / 2
+
+        # self.mode_vec = np.conjugate(self.mode_vec)
```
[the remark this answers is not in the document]
fakufaku: Definitely!

pyroomacoustics/doa/srp.py
fakufaku: @Womac is indeed correct. However, keeping the offset may help when plotting the cost function. I think that in SRP-PHAT all the diagonal components should be one, and in that case, the constant offset is always the number of channels. I will check this again though. But in that case, we could just add this so that the cost function is always positive, which is not really necessary, but may be useful sometimes.

Womac: @fakufaku, just an idea,  if we want to keep the diagonal, I think we can write this even shorter, couldn't we just `srp_f = np.real(np.conj(mode_vec[..., None, :])@ CC @ mode_vec[..., None])` where `np.real` is only there to avoid imaginary parts due to rounding errors. Afterwards, we could just sum over all frequencies to obtain the global estimate. If we store the grid_points in one of the first dimensions of mode_vec and put a dimension of one in CC there, we could also avoid the for-loop, then the shapes could be something like: `mode_vec.shape = (grid_points, frequencies, M)`, and `CC.shape = (1,frequencies, M, M)` In the end with `SRP = np.sum(srp_f, axis=1)` we get the SRP coefficients [not normalised yet], where `SRP.shape = (grid_points, 1,1)`? Finally, `spr_cost = SRP/ self.num_snap / self.num_freq / self.num_pairs` What do you think?

fakufaku: @ebezzam I have added more comments. Is it easier to parse now ?

pyroomacoustics/doa/srp.py
```diff
@@ -63,24 +83,54 @@ def _process(self, X):
...
         for k in self.freq_bins:
-            CC.append( np.dot(pX[:,k,:], np.conj(pX[:,k,:]).T) )
+            CC.append(np.dot(pX[:, k, :], np.conj(pX[:, k, :]).T))
         CC = np.array(CC)
 
+        M = self.L.shape[1]
+        ar = np.arange(M)
+        av = ar[:, None]
+
+        # the two masks here allow to select the upper triangular part
```
ebezzam, line 93: sorry for nitpicking, there's only one mask left now

pyroomacoustics/doa/srp.py
ebezzam: @fakufaku yes really easy now!

pyroomacoustics/doa/srp.py
fakufaku: 😆 fixed

ebezzam (approved):

PR #211, opened 2021-03-15
MUSIC Implementation: Use normalization

4bian:
Hi,
I recommend to use a normalization step to improve the localization performance of MUSIC:

Previous MUSIC implementation:
I. Calculate MUSIC pseudo spectrum for each frequency bin
II. Average the MUSIC pseudo spectra across the frequency axis
III. Peak detection

Comment: Averaging without normalization is suboptimal, as the maxima of each frequency bin-wise MUSIC pseudo spectrum can differ in orders of magnitude.

Recommended MUSIC implementation:
I. Calculate MUSIC pseudo spectrum for each frequency bin
II.a. Normalize the MUSIC pseudo spectrum for each frequency bin
II.b. Average the normalized MUSIC pseudo spectra across the frequency axis
III. Peak detection

There is also a paper from 2014, where this issue was investigated [1]: The implementation with normalization constantly outperformed the non-normalized approach. 

Best Regards,
Fabian

[1] D. Salvati, C. Drioli and G. L. Foresti, "Incoherent Frequency Fusion for Broadband Steered Response Power Algorithms in Noisy Environments," in IEEE Signal Processing Letters, vol. 21, no. 5, pp. 581-585, May 2014, doi: 0.1109/LSP.2014.2311164.


pyroomacoustics/doa/music.py
```diff
@@ -84,7 +89,15 @@ def _process(self, X):
...
         self.Pssl = self._compute_spatial_spectrumvec(cross)
+        if self.frequency_normalization:
+            self._apply_frequency_normalization()
         self.grid.set_values(np.squeeze(np.sum(self.Pssl, axis=1) / self.num_freq))
+        
+    def _apply_frequency_normalization(self):
+        """
+        Normalize the MUSIC pseudo-spectrum per frequency bin
+        """
+        self.Pssl = self.Pssl / np.max(self.Pssl, axis=0)[None,...]
```
fakufaku, line 100: This could be written slightly more explicitly as `np.max(self.Pssl, axis=0, keepdims=True)`.

fakufaku (commented):
Thank you so much for the contribution! 🤗 I'd be very happy to merge this, but would it be possible to make a few changes ?

1. Set the default `frequency_normalization=False` so that the behavior of existing code does not change.
2. Add the reference to the paper in the docstring of MUSIC.

pyroomacoustics/doa/music.py
4bian: Hi, thank you for the comments.
I totally agree with 2. and the proposed modification in line 100.

However, I also see a good reason to use frequency_normalization=True as the default behavior.

Obviously, the advantage of frequency_normalization=False is backward compatibility.

From a user's perspective, I think frequency_normalization=True is beneficial. Often, users don't tweak every parameter of a library and just use the standard parameters. Considering that, I think frequency_normalization=True is advantageous due to its much better performance.

fakufaku: I understand where you are coming from! From my point of view, there are two important reasons to keep the default behavior to `frequency_normalization=False`

1. Backward compatibility
2. This is the *expected* implementation of MUSIC, i.e., how it is described in most textbooks (that I am aware of)

What about creating a new thin-wrapper class around the `MUSIC` class, e.g., `NormMUSIC`, where normalization is active by default ? I think this would be a good middle ground solution. By having an extra class, the existence of this option would be prominently featured in the doc. Then, the `MUSIC` class would still be backward compatible and do what one expect by default.

To further make the point, what about creating an example script comparing `MUSIC` to `NormMUSIC` in terms of accuracy ? Then, we can point to it in the doc so that people use `NormMUSIC` first.

4bian: Hi, thanks you for your suggestion and sorry for the late answer.
I agree with you, an extra class "NormMUSIC" is the better solution. 
I will implement it soon.

fakufaku: Fantastic! Looking forward to it! 👍

fakufaku:
Hi @4bian , sorry I dropped the ball on this PR! But I think it is looking good now and I could merge it as soon as possible. Could you please just merge with master and resolve the conflict in `doa/__init__.py` ? Then I can directly merge it into master. If you can't do it  I could also do it myself, as you prefer!

4bian:
I rebased the PR branch to get the latest commits from the base branch.

fakufaku:
Thanks for rebasing! I'm sorry, there is one more thing... We use the [black](https://github.com/psf/black) code formatter to have uniform fomatting of the code base. Could you please apply it and re-commit ? All you need to do is
```bash
pip install black
black ./pyroomacoustics
```
(assuming you are already in the pyroomacoustics root folder)
Thanks a lot!

4bian:
Hi, I used black to reformat the code. 
However, I'm a bit confused, as black also modified files that I didn't touch.

fakufaku:
Merged in #242

PR #215, opened 2021-04-12
Spherical to cartesian conversion and prettier RIR plot

fakufaku:
- A function to convert from spherical to cartesian coordinates in the DOA subpackage. The reverse function already existed.
- Cleans up the plot_rir function in Room so that the labels are neater. It also changes the plot with the `FD` option to display spectrograms of the RIR.

- [X] Docstrings
- [ ] Tests (we'll see in CI)
- [ ] Doc build (we'll see in CI)
- [X] Unit test
- [X] CHANGELOG


pyroomacoustics/doa/utils.py
```diff
@@ -33,20 +34,52 @@ def great_circ_dist(r, colatitude1, azimuth1, colatitude2, azimuth2):
...
+            )
+            ** 2
+        ),
+        np.cos(colatitude1) * np.cos(colatitude2)
+        + np.sin(colatitude1) * np.sin(colatitude2) * np.cos(d_azimuth),
+    )
     return dist
 
+
+def cart2spher(vectors):
```
ebezzam, line 59: Nice! actually this will be useful for the PR #228!

pyroomacoustics/room.py
```diff
@@ -523,9 +523,19 @@ def callback_mix(premix, snr=0, sir=0, ref_mic=0, n_src=None, n_tgt=None):
 def wall_factory(corners, absorption, scattering, name=""):
     """ Call the correct method according to wall dimension """
     if corners.shape[0] == 3:
-        return Wall(corners, absorption, scattering, name,)
+        return Wall(
+            corners,
+            absorption,
+            scattering,
+            name,
+        )
```
ebezzam, line 531: Regarding this formatting edits, I guess there's a bunch of places that could use such fixes. Should we run something like [black](https://github.com/psf/black) across the whole package?

pyroomacoustics/room.py
```diff
@@ -1530,26 +1548,69 @@ def plot_rir(self, select=None, FD=False):
...
+
+                if FD:
+                    axes[k,0].set_ylabel("Freq. [kHz]")
+
+                if k < len(pairs) - 1:
+                    axes[k,0].set_xticks([])
+                else:
+                    axes[k,0].set_xlabel("Time [ms]")
+            fig.align_ylabels(axes[:, 0])
+            fig.tight_layout()
```
ebezzam, line 1613: Should we return `fig` and `axes` in case the user would like to make additional changes?

ebezzam (changes requested):
Sorry for the very late feedback! Looks good, one possible change for the `plot_rir`.

PR #217, opened 2021-04-12
Only enable ray tracing if desired.

hrosseel:
Bugfix in room.py: only enable ray tracing if desired.

### DIFF
--- a/pyroomacoustics/room.py
+++ b/pyroomacoustics/room.py
@@ -775,9 +775,8 @@ class Room(object):
             self.set_air_absorption()
 
         # default values for ray tracing parameters
-        self.set_ray_tracing()
-        if not ray_tracing:
-            self.unset_ray_tracing()
+        if ray_tracing:
+            self.set_ray_tracing()
 
         # in the beginning, nothing has been
         self.visibility = None


fakufaku:
Hi @hrosseel , thanks for taking the time do a PR!
Do you have some code example where the bug occurs, i.e., you set `ray_tracing=False`, but it is actually enabled ?

The rational behind the original code is that calling once `self.set_ray_tracing()` will set default values for a number of parameters.
Then, if ray tracing is not needed, i.e., `ray_tracing=False`, we call `self.unset_ray_tracing()` to deactivate it.
```python
# default values for ray tracing parameters
self.set_ray_tracing()
if not ray_tracing:
    self.unset_ray_tracing()
```
I just checked by running the `room_from_rtl60.py` example with `-m ism` argument, which runs only the image source model.
I could check that `room.simulator_state["rt_needed"]`, which controls whether ray tracing is used or not, was set to `False`. In addition, `room.simulator_state["rt_done"]` is also set to `False`, which means the ray tracing code has not been called.

There may be some corner case where the above sequence fails and an incorrect behavior happens, so your feedback on the matter is very appreciated!

hrosseel:
Hi @fakufaku, thanks for having a look at my PR. 

I was not aware that the `self.set_ray_tracing()` function initializes other parameters. I made this PR request following a warning (`n_rays > 100000`) which got invoked when I wasn't using any ray tracing. Which I thought was a bug.

To me, initializing parameters which you are eventually not going to need, seems kind of redundant. Why would you need to initialize ray tracing parameters you are not using? :) 

In case you are initializing other parameters in `self.set_ray_tracing()`, isn't there a better place to initialize them? E.g. at a higher level?

Thanks for your feedback.

fakufaku:
I can't remember for sure why I did it this way. You are right that the warning is very confusing if you did not specify `ray_tracing=True` and this qualifies as a bug :)
I think I did the default initialization because having non-initialized values triggered some other problems later, although I don't remember for sure. I will try to see if it possible to get rid of the warning when `ray_tracing=False`.
The initialization here is already at a fairly high level (Room constructor), so I'm not sure where would be more appropriate, but if you have ideas let me know!

fakufaku:
@hrosseel It has been quite a long time, but the issue has been fixed as part of PR #237. Now there should not be a warning anymore when ray tracing is not used!
Thanks for your patience!

PR #230, opened 2021-07-21
fix wrong bracketing for negative values in is_inside

Nimo139:
In the Shoebox model, negative values are considered to be in the room. 

Minimal example:

```
import numpy as np
import pyroomacoustics as pra
# Create a 4 by 6 metres shoe box room
room = pra.ShoeBox([4,6])
# Add a source somewhere in the room
room.add_source([2.5, 4.5])
print(room.is_inside([1.0, 3.0]))  # True
print(room.is_inside([1.0, 6.1]))  # False
print(room.is_inside([-1.0, 3.0]))  # True, false is expected
```

Happy PR :smiley:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,7 @@ adheres to `Semantic Versioning <http://semver.org/spec/v2.0.0.html>`_.
 `Unreleased`_
 -------------
 
-Nothing yet.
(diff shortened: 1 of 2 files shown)


fakufaku:
Wow! Thank you so much for the catch! That was a hideous bug 🙈  The PR goes right in!

PR #250, opened 2022-02-10
Randomized image method

orchidas:
- Option for using randomized image method for removing sweeping echoes in shoebox rooms in room.py. Use flag `use_rand_ism=True`.
- Sweeping echo measure in metrics.py
- Test and example in tests/test_random_ism.py
- Updated CHANGELOG

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,7 @@ Added
 
 - New DOA method: MUSIC with pseudo-spectra normalization. Thanks @4bian!
   Normalizes MUSIC pseudo spectra before averaging across frequency axis.
+- Randomised image source method for removing sweeping echoes in shoebox rooms.
 
 Bugfix
 ~~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,7 +30,7 @@ Added
 
 - New DOA method: MUSIC with pseudo-spectra normalization. Thanks @4bian!
   Normalizes MUSIC pseudo spectra before averaging across frequency axis.
-- Randomised image source method for removing sweeping echoes in shoebox rooms.
+- Randomized image source method for removing sweeping echoes in shoebox rooms.
 
 Bugfix
 ~~~~~~
--- a/examples/randomized_image_method.py
+++ b/examples/randomized_image_method.py
@@ -50,8 +50,7 @@ ax.set_ylabel('Y')
 ax.set_zlabel('Z')
 
 plt.xlim([0,4])
-plt.ylim([0,4])
-plt.show()    
+plt.ylim([0,4])  
 
 
 ###############################################################
@@ -102,7 +101,6 @@ for n in range(M):
         #plot RIR
         plt.figure()
         plt.plot(room.rir[n][0])
-        plt.show()
         
         
     # measure of spectral flatness of sweeping echos
--- a/examples/randomized_image_method.py
+++ b/examples/randomized_image_method.py
@@ -1,0 +1,162 @@
+"""
+Randomized image method example
+===============================
+
+In this example, we will show the benefits of using the randomized
+image method to remove sweeping echoes in RIRs simulated with ISM.
+"""
+
+import numpy as np
+import pyroomacoustics as pra
+import matplotlib.pyplot as plt
+from pyroomacoustics.transform import stft
+from pyroomacoustics import metrics as met
+from mpl_toolkits.mplot3d import Axes3D
+
+
+# create an example with sweeping echo - from Enzo's paper
+room_size = [4, 4, 4]
+source_loc = [1,2,2]
+center_x = room_size[0]/2
+center_y = room_size[1]/2
+center_z = room_size[2]/2
+
+# number of mics
+M = 15
+mic_loc = np.zeros((3,M), dtype = float)
+gridSpacing = 0.25
+
+
+
+# =============================================================================
+# create uniform grid of mics
+# =============================================================================
+count = 0
+Nmic_each_axis = int((M/3 - 1)/2)
+for n in range(-Nmic_each_axis, Nmic_each_axis+1):
+    mic_loc[:,count] = [center_x+n*gridSpacing, center_y, center_z]
+    mic_loc[:,count+1] = [center_x, center_y+n*gridSpacing, center_z]
+    mic_loc[:,count+2] = [center_x, center_y, center_z+n*gridSpacing]
+    count += 3
+              
+
+fig = plt.figure()
+ax = fig.add_subplot(111, projection='3d')
+
+ax.scatter(mic_loc[0,:],mic_loc[1,:], mic_loc[2,:])
+
+ax.set_xlabel('X')
+ax.set_ylabel('Y')
+ax.set_zlabel('Z')
+
+plt.xlim([0,4])
+plt.ylim([0,4])
+plt.show()    
+
+
+###############################################################
+# normal ISM with sweeping echoes
+###############################################################
+
+# sampling frequency
+fs = 44100
+
+room = pra.ShoeBox(room_size, fs, materials=pra.Material(0.1), max_order = 50)
+
+room.add_source(source_loc)
+
+room.add_microphone_array(pra.MicrophoneArray(mic_loc, fs))
+
+room.compute_rir()
+
+#plot spectrograms to check for sweeping echoes
+
+fft_size = 512  # fft size for analysis
+fft_hop = 128  # hop between analysis frame
+fft_zp = 512  # zero padding
+analysis_window = pra.hann(fft_size)
+
+print("Sweeping echo measure for ISM is :")
+for n in range(M):
+    
+    if n == 0:
+        S = stft.analysis(room.rir[n][0],  fft_size, fft_hop, win=analysis_window, zp_back=fft_zp)
+        
+        plt.figure()
+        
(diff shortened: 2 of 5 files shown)


pyroomacoustics/room.py
```diff
@@ -769,6 +769,7 @@ def __init__(
         humidity=None,
         air_absorption=False,
         ray_tracing=False,
+        use_rand_ism=False,
```
fakufaku, line 772: Can you please add documentation for the new parameter in the docstring ?

I think it would be great if we can add a section on the method in the documentation at the top of the file. We could even add a figure there.

pyroomacoustics/tests/test_random_ism.py
```diff
@@ -0,0 +1,176 @@
+
+"""
+Created on Tue Feb  8 17:50:45 2022
+
+@author: od0014
+
+Script to test removal of sweeping echoes with randomized image method
```
fakufaku, line 7: The current test file is not suitable for automatic testing. Functions with name prefixed by `test_` are automatically run at test time by the CI system. They should not take any argument. The testing inside is typically done by using `assert` on some condition that is `True` if the code being tested is working as expected and `False` otherwise.

fakufaku (changes requested):
@orchidas Thank you so much for the contribution! Sorry it took me a little time to get to the review. There are a few things to fix before we can merge. These are in the comments below, but I summarize them here as well.
1. Document the new parameter in the docstring of the functions where it appears
2. Change the test script so that it can run automatically (no figures, use assert to check, no parameters)
3. [would be nice] Add a section/paragraph describing the randomized ISM in the documentation at the top of `room.py`.

orchidas:
All requested changes have been added. No plot or print statements in test, new file in examples folder and docstrings updated in room.py.

fakufaku:
@orchidas Thanks for the update! I think we are almost there! If you could just do the following modifications that would be fantastic.
1. The test do not pass due to the function `test_sweep_measure`. Pytest/nose will just run all the functions/files that have `test_` or a variation thereof in the name. These functions should not take an mandatory arguments. This one is not a test function as far as I can tell, so could you please rename it to something that does not trigger a test, e.g., `measure_sweeping_echo`, or just call `ssf = met.sweeping_echo_measure(rir,fs)` directly from the other functions.
2. Run [black](https://github.com/psf/black) to make the formatting consistent (version >= 22)
```bash
pip install black
black pyroomacoustics
```
3. In the example file, it would be great if you can remove all the intermediate `plt.show()` and just put one at the end, so that we can see all the figures at once. Also, it would make the result much easier to understand if you add some titles to the figures, e.g. "without random ISM", "with random ISM", etc.

For my own education, in the example, we can still some sweeping echo in the lower frequencies when using the random ISM. Would this disappear if we use a larger displacement ?
Also, the amplitude in the regular ISM is much larger than the random one. Is it due to the all the echoes arriving at the same time ?

Cheers!

fakufaku:
This is looking pretty good, all the tests are passing! There is just the black formatting that seems broken. It doesn't show in the test due to some misconfiguration on my part. I think the standard formatting has changed recently so you should make sure you have the latest version of black before running (>v22).
Once this is done I will merge this in the main branch!

orchidas:
Hi Robin, I downloaded the latest version of black and re-did the formatting. I got this error error: cannot format pyroomacoustics/libroom_src/ext/eigen/scripts/relicense.py: Cannot parse: 57:18:             print "SKIPPED", filename
Python 2 support was removed in version 22.0.

Regarding your other questions, I am in touch with Enzo and will get back to you shortly.

CHANGELOG.rst
```diff
@@ -30,6 +30,7 @@ Added
 
 - New DOA method: MUSIC with pseudo-spectra normalization. Thanks @4bian!
   Normalizes MUSIC pseudo spectra before averaging across frequency axis.
+- Randomised image source method for removing sweeping echoes in shoebox rooms.
```
fakufaku, line 33: Very, *very* minor comment, the US spelling is used in other locations 😄

examples/randomized_image_method.py
```diff
@@ -0,0 +1,168 @@
...
+        )
+        ax.set_title("RIR for Mic location " + str(n) + " with random ISM")
+        ax.set_ylabel("Frequency")
+        ax.set_xlabel("Time")
+        ax.set_aspect("auto")
+        ax.axis("off")
+     
+        
+        #plot RIR
+        plt.figure()
```
fakufaku, line 156: Here I would add a title to the RIR to match it to the corresponding spectrogram. Or put both in a subplots of a single figure.

pyroomacoustics/metrics.py
```diff
@@ -260,3 +261,94 @@ def pesq(ref_file, deg_files, Fs=8000, swap=False, wb=False, bin="./bin/pesq"):
...
+    A higher value indicates less sweeping echoes
+
+    For details see : De Sena et al. "On the modeling of rectangular geometries in
+    room acoustic simulations", IEEE TASLP, 2015
+
+
+
+    Parameters
+    ----------
+    rir : RIR signal from ISM (mono).
```
fakufaku, line 278: The description should be one line down and indented.

fakufaku: Same for `fs`

pyroomacoustics/room.py
```diff
@@ -98,6 +98,37 @@
...
     actually **not** the energy absorption we use now.  The ``absorption``
     parameter is now deprecated and will be removed in the future.
+    
+    
+Randomized Image Method
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In highly symmetric shoebox rooms, the regularity of image sources’ positions 
+leads to a monotonic convergence in the time arrival of far-field image pairs.
+This causessweeping echoes. The randomized image method adds a small random
```
fakufaku, line 108: `causessweeping` -> `causes sweeping`

pyroomacoustics/room.py
```diff
@@ -98,6 +98,37 @@
...
+
+In highly symmetric shoebox rooms, the regularity of image sources’ positions 
+leads to a monotonic convergence in the time arrival of far-field image pairs.
+This causessweeping echoes. The randomized image method adds a small random 
+displacement to the image source positions, so that they are no longer 
+time-aligned, thus reducing sweeping echoes considerably.
+
+To use the randomized method, set the flag ``use_rand_ism`` to True while creating
+a room. Additionally, the maximum displacement of the image sources can be 
+chosen by setting the parameter ``max_rand_disp``. The default value is 8cm.
```
fakufaku, line 114: You could reference the example here, e.g., For a full example see `examples/randomized_image_method.py`.

pyroomacoustics/room.py
```diff
@@ -98,6 +98,37 @@
...
+    room_dim = [5, 5, 5]  # meters
+
+    # We invert Sabine's formula to obtain the parameters for the ISM simulator
+    e_absorption, max_order = pra.inverse_sabine(rt60, room_dim)
+
+    # Create the room
+    room = pra.ShoeBox(
+        room_dim, fs=16000, materials=pra.Material(e_absorption), max_order=max_order,
+        use_rand_ism = True, max_rand_disp = 0.05
+        )
```
fakufaku, line 131: The indent of the closing parenthesis is not consistent.

pyroomacoustics/room.py
```diff
@@ -1941,6 +1987,23 @@ def image_source_model(self):
...
+                # if randomized image method is selected, add a small random
+                # displacement to the image sources
+                if self.simulator_state["random_ism_needed"]:
+
+                    nImages = np.shape(source.images)[1]
+
+                    # maximum allowed displacement is 8cm
+                    max_disp = self.max_rand_disp
+                    # add a random displacement to each cartesian coordinate
+                    disp_x = np.random.uniform(-max_disp, max_disp, nImages)
```
fakufaku, line 1999: I think this could be written more compactly
```python
disp = np.random.uniform(-max_disp, max_disp, size=(3, n_images)
source.images += disp
```

pyroomacoustics/room.py
```diff
@@ -1941,6 +1987,23 @@ def image_source_model(self):
                 source.damping = self.room_engine.attenuations.copy()
                 source.generators = -np.ones(source.walls.shape)
 
+                # if randomized image method is selected, add a small random
+                # displacement to the image sources
+                if self.simulator_state["random_ism_needed"]:
+
+                    nImages = np.shape(source.images)[1]
```
fakufaku, line 1994: Also pretty minor but we use snake_case for variables, so `n_images`.

pyroomacoustics/room.py
```diff
@@ -2597,6 +2660,12 @@ class ShoeBox(Room):
     ray_tracing: bool, optional
         If set to True, the ray tracing simulator will be used along with
         image source model.
+    use_rand_ism: bool,optional
```
fakufaku, line 2663: space missing -> `bool, optional`

fakufaku (changes requested):
I found a few minor things to correct. After that we should be good to go.

fakufaku:
Hi @orchidas , thank you very much. You do not have to worry about the file `pyroomacoustics/libroom_src/ext/eigen/scripts/relicense.py`, it part of a dependency, not the pyroomacoustics codebase.

fakufaku:
Finally merged! Sorry it took so long!

PR #259, opened 2022-05-17
DIRPAT RIR Code First Commit Review

prerak23:

Happy PR :smiley:


fakufaku:
Hi @prerak23 ! Thank you so much for the PR! I will review it in the coming days! 👍

fakufaku:
@prerak23 It looks like the file `open_sofa_interpolate` is missing. See the test output [here](https://github.com/LCAV/pyroomacoustics/runs/6534777751?check_suite_focus=true#step:8:48)


```
pyroomacoustics/directivities.py:9: in <module>
    import open_sofa_interpolate
E   ModuleNotFoundError: No module named 'open_sofa_interpolate'
```

prerak23:
Thanks, i will make a note of all the corrections given as the review of the pull request goes ahead .
But yes the DIRPAT sofa files are not present in the github repo.
You can download them from here.
https://phaidra.kug.ac.at/detail/o:68229#?q=DIRPAT&page=1&pagesize=10

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+            self.Ysh_tilda_inv, self.w_ = calculation_pinv_voronoi_cells(self.Ysh, self.theta_16, self.no_of_lat)
+
+
+            self.freq_angles_fft = np.zeros((self.no_of_nodes, self.samples_size_ir), dtype=np.complex_) #N-point FFT
+            self.sh_coeffs_expanded_original_grid = 0
+            self.sh_coeffs_expanded_target_grid = 0
+
+            # Rotate the target grid (Fibonacci grid)
+
+            self.rotated_fibo_points = np.matmul(res, self.points.T)
```
fakufaku, line 236: I see that you are currently rotating all the points of the interpolated grid.
Another solution is to rotate the incident points when you measure the directivity.
This would make it easier to share the interpolated grid in case multiple objects using the same directivity are used without having to recompute the interpolation for all of them.

Actually, I am now thinking that we could just store the interpolated grids with pyroomacoustics instead of loading from DIRPAT and interpolating every time... Thoughts ?

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+            if self.source == True:
+                file_sofa = sofa.Database.open(path)
+
+                # Receiver positions
+
+                rcv_pos = file_sofa.Receiver.Position.get_values()
+
+                print(rcv_pos.shape)
+
+                # CHEAP HACCCKK SPECIFICALLY FOR DIRPAT
```
fakufaku, line 327: Will this fail if we use this code for general sofa files ? How difficult would it be to make it generic ?

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
+import numpy as np
+import scipy
+import sofa
```
fakufaku, line 3: What happens if `sofa` is not installed ? Have you added it to the list of dependencies ?

I like to keep the required dependencies to a minimum, so can we do something like
```python
try:
     import sofa
     has_sofa = True
except ImportError:
     has_sofa = False

def function_using_sofa():
    if not has_sofa:
        raise ValueError("The package 'sofa' needs to be installed to call this function. Install by doing `pip install sofa`")
````
```

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+            self.cal_spherical_coeffs_grid_and_interpolate()
+
+
+        else:
+
+            self.freq_angles_fft = np.zeros((self.no_of_nodes, self.samples_size_ir), dtype=np.complex_)
+            for i in range(self.no_of_nodes):
+                self.freq_angles_fft[i, :] = fft(self.sofa_msr_fir[i, :])
+            '''
+            n_c = (self.colatitude_simulation)
```
fakufaku, line 251: remove commented out lines

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+
+    def cal_spherical_coeffs_grid_and_interpolate(self):
+
+        """
+        Weighted least square solution to calculate discrete sphreical harmonic coeffs , Using the spherical harmonic coeffs
+        interpolation is done on fibonacci sphere
+        """
+
+
+        #freq_angles_fft = np.zeros((self.no_of_nodes, self.samples_size_ir// 2), dtype=np.complex_)
```
fakufaku, line 283: remove commented out line

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+                IR_S = file_sofa.Data.IR.get_values()
+
+                # Look for source of specific type requested by user"
+
+                rcv_msr = IR_S[self.id, :, :]  # First receiver #Shape ( no_sources * no_measurement_points * no_samples_IR)
+
+                # downsample the fir filter.
+                if "Debug_sofa_file" not in self.path:
+                    rcv_msr = decimate(rcv_msr, int(round(file_sofa.Data.SamplingRate.get_values()[0] / fs)), axis=-1)
+                    print("Decimating FIR'S Source")
```
fakufaku, line 346: We probably want to remove print statements. If you want to raise a warning, use the `warning` package, if an error, raise an exception.

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+
+                no_of_nodes = 540
+
+                # samples = file_sofa.Dimensions.N  # Samples per IR
```
fakufaku, line 350: remove commented out

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+
+
+                return phi_rcv, theta_rcv, r_rcv, theta_16, rcv_msr, no_of_nodes, samples, no_of_lat
+            else:
+                file_sofa = sofa.Database.open(path)
+                # Source positions
+
+                src_pos = file_sofa.Source.Position.get_values()
+
+                # CHEAP HACCCKK SPECIFICALLY FOR DIRPAT
```
fakufaku, line 380: can we get rid of cheap hacks easily ?

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+
+                IR_S = file_sofa.Data.IR.get_values()
+
+                # Look for receiver of specific type requested by user"
+
+                rcv_msr = IR_S[:, self.id,:]  # First receiver #Shape (no_measurement_points * no_receivers * no_samples_IR)
+
+                if "Debug_sofa_file" not in self.path:
+                    rcv_msr = decimate(rcv_msr, int(round(file_sofa.Data.SamplingRate.get_values()[0] / fs)), axis=-1)
+                    print("Decimating FIR'S Receiver")
```
fakufaku, line 398: remove print statement

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+        """
+
+        if self.interpolate:
+            #dd, ii = self.nn_kd_tree_rotated_fibo_grid.query([longitude, latitude])  #Query on the rotated set of points
+            return self.sh_coeffs_expanded_target_grid[index, :]
+        else:
+            #dd, ii = self.nn_kd_tree_rotated_sofa_grid.query([longitude, latitude])  # Query on the rotated set of points
+            return self.sofa_msr_fir[index, :]
+
+    def change_orientation(self, azimuth_change, colatitude_change):
```
fakufaku, line 470: I think this is why I suggested to do the rotation when computing the response for directions rather than on the grid itself.

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+        self.phi_sofa, self.theta_sofa, self.r_sofa, self.theta_16, self.sofa_msr_fir, self.no_of_nodes, self.samples_size_ir, self.no_of_lat = self.open_sofa_database(path=self.path,
+            fs=self.fs)
+
+        self.sofa_x,self.sofa_y,self.sofa_z=sph2cart(self.phi_sofa,self.theta_sofa,self.r_sofa)
+
+        self.interpolate = interpolate
+
+        #Rotation matrix
+
+        n_c = (colatitude_simulation)
```
fakufaku, line 199: The rotation here assumes that the original response is pointing to [1, 0, 0]. This is fine, but should be documented in the docstring.

Also, there is no rotation wrt the x-axis. This is fine for symmetric responses, such as cardioid, but probably not for measured ones. To be general, we probably want to provide a full rotation with 3 angles.

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+
+        : Return : nd.array() return nearest index on the grid based on the input angles using kDtree
+        """
+
+        longitude=azimuth.reshape(-1,1)
+        latitude=colatitude.reshape(-1,1)
+
+        start=timer()
+        if self.interpolate:
+            dd,ii=self.nn_kd_tree_rotated_fibo_grid.query(np.hstack((longitude,latitude)))
```
fakufaku, line 440: So my proposal of apply rotation to query point would be to rotate longitude/latitude here. This would be easier to do if the grid is stored in cartesian coordinates.

Also, just a detail, but why add longitude/latitude notation here ? Maybe should stick to azimuth/coltitude throughout.

pyroomacoustics/open_sofa_interpolate.py
fakufaku: Actully, as I wrote above, we need to handle arbitrary rotations with three angles. Also, this code assumes the starting point is at [1, 0, 0]. If you apply it twice, the result might not be as expected.
That is
```python
obj2 = obj  # same starting point

obj.change_orientation(az1, col1)
obj.change_orientation(az2, col2)
# is different than
obj2.change_orientation(az1 + az2, col1 + col2)
```

fakufaku: This is easy to see if you recall that matrices do not commute.

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
+import numpy as np
+import scipy
+import sofa
+from scipy.fft import fft,fftfreq,ifft
+import math
+import matplotlib.pyplot as plt
```
fakufaku, line 6: Same here because `matplotlib` is not a hard requirement for pyroomacoustics. Please protect it as above.

pyroomacoustics/open_sofa_interpolate.py
```diff
@@ -0,0 +1,556 @@
...
+        """
+        Plot 3D plots of the SOFA file , interpolated fibonnacci sphere and roatated fibo and sofa spheres.
+
+        Display gain for a specified freq bin
+
+
+        :Parameter (int) freq_bin:Bin number
+
+        """
+
```
fakufaku, line 505: To protect matplotlib import, one solution is to import inside the function. Or use a `has_matplotlib` flag set when doing the initial import.

pyroomacoustics/room.py
```diff
@@ -617,13 +583,27 @@ def callback_mix(premix, snr=0, sir=0, ref_mic=0, n_src=None, n_tgt=None):
...
 from .utilities import angle_function
+from .directivities import CardioidFamily, source_angle_shoebox, Directivity, DIRPATRir
+import matplotlib.pyplot as plt
+from math import floor
+from timeit import default_timer as timer
+from scipy.fft import fft, fftfreq, ifft
+from scipy.signal import fftconvolve
+import sys
+
+sys.path.insert(0, "/home/psrivast/PycharmProjects/axis_2_phd")
```
fakufaku, line 599: This should be changed to proper import of package file.

pyroomacoustics/room.py
```diff
@@ -617,13 +583,27 @@ def callback_mix(premix, snr=0, sir=0, ref_mic=0, n_src=None, n_tgt=None):
...
+# import open_sofa_interpolate
 
+from scipy import fft
+from scipy.signal import hilbert
+
+
+# from ..pyroomacoustics.axis_2phd.3d_subplot import interpolate_sofa_files
```
fakufaku, line 606: please remove this line.

pyroomacoustics/room.py
```diff
@@ -904,14 +871,23 @@ def _var_init(
...
         # in the beginning, nothing has been
         self.visibility = None
 
         # initialize the attribute for the impulse responses
         self.rir = None
 
+        self.sh_deg = 12
+        self.print_filter = 0
+
+        self.DIRPAT_source_files=["LSPs_HATS_GuitarCabinets_Akustikmessplatz.sofa","Debug_sofa_file_source_2.sofa"]
```
fakufaku, line 887: Maybe it would be better to have these variables imported from a dirpat specific file.

pyroomacoustics/room.py
```diff
@@ -2084,143 +2040,419 @@ def compute_rir(self):
...
+            The constructed RIR still lacks air absorption and distance absorption because in the old pyroom these calculation happens on the octave band level.
+
+
+        """
+
+        attenuations = attenuations / dist
+        alp = []
+        window_length = 81
+
+        lut_pos = np.arange(-32, 32, 1)
```
fakufaku, line 2314: It looks like this is not used. Can we remove it ?

pyroomacoustics/room.py
```diff
@@ -2084,143 +2040,419 @@ def compute_rir(self):
...
+            Constructed RIR for this particlar src mic pair .
+
+            The constructed RIR still lacks air absorption and distance absorption because in the old pyroom these calculation happens on the octave band level.
+
+
+        """
+
+        attenuations = attenuations / dist
+        alp = []
+        window_length = 81
```
fakufaku, line 2312: The window length can be obtained from the pyroomacoustics internal parameters
```python
from . import constants
window_length = constants.get("frac_delay_length")
```

pyroomacoustics/room.py
```diff
@@ -2084,143 +2040,419 @@ def compute_rir(self):
...
+                final_fir_IS_len = (fir_length_octave_band + window_length +
+                                self.sources[src_pos].directivity.obj_open_sofa_inter.samples_size_ir) - 2
+                ind_knn_src = self.sources[src_pos].directivity.obj_open_sofa_inter.cal_index_knn(azi_s, col_s)
+            elif rec_presence:
+                final_fir_IS_len = (fir_length_octave_band + window_length +
+                                self.mic_array.directivity[mic_pos].obj_open_sofa_inter.samples_size_ir) - 2
+                ind_knn_mic = self.mic_array.directivity[mic_pos].obj_open_sofa_inter.cal_index_knn(azi_m, col_m)
+
+       
+        #else:
```
fakufaku, line 2342: remove

pyroomacoustics/room.py
```diff
@@ -2084,143 +2040,419 @@ def compute_rir(self):
...
+        cpy_ir_len_4 = np.zeros((no_imag_src,final_fir_IS_len), dtype=np.complex_)
+        att_in_dft_scale = np.zeros((no_imag_src,fir_length_octave_band), dtype=np.complex_)
+
+        for i in range(no_imag_src):  # Loop through Image source
+
+            att_in_octave_band = attenuations[:, i]
+            att_in_dft_scale_=att_in_dft_scale[i,:]
+
+            # Interpolating attenuations given in the single octave band to a DFT scale.
+            #for b, bw in enumerate(bws):
```
fakufaku, line 2370: remove

pyroomacoustics/room.py
```diff
@@ -2084,143 +2040,419 @@ def compute_rir(self):
...
+
+            att_in_dft_scale_ = self.octave_bands.octave_band_dft_interpolation(att_in_octave_band, self.air_absorption, dist[i], att_in_dft_scale_,bws, self.min_phase)
+
+
+            # Minmum phase...
+            # k+=1e-07
+            #if self.min_phase:
```
fakufaku, line 2377: If not used, remove

pyroomacoustics/room.py
```diff
@@ -2084,143 +2040,419 @@ def compute_rir(self):
...
+                out = a * b
+
+                out = ifft(out)
+
+                ir_diff[time_ip:(time_ip + final_fir_IS_len)] += np.real(out)
+            '''
+
+
+        end=timer()
+        print("Image source time taken",end-start)
```
fakufaku, line 2443: remove print statement

pyroomacoustics/room.py
```diff
@@ -787,29 +765,22 @@ class Room(object):
...
+            t0=0.0,
+            max_order=1,
+            sigma2_awgn=None,
+            sources=None,
+            mics=None,
+            temperature=None,
+            humidity=None,
+            air_absorption=False,
+            ray_tracing=False,
+            min_phase=False,
```
fakufaku, line 783: Please document new argument in function docstring.

fakufaku: Also, is this only used for the DIRPAT responses ? If I understand correctly this can be use for the band-pass filters also in regular use of pyroomacoustics. Is it taken into account in that case too ?

pyroomacoustics/test_pra.py
```diff
@@ -0,0 +1,591 @@
+import matplotlib
+
```
fakufaku, line 2: The test files should be located in `pyroomacoustics/tests` and should be able to run automatically. Is it the case ?

They should be functions with a name that starts with `test_`, e.g., `test_dirpat_stuff`, take no argument, and return nothing. The test success/failure is determined using assert.

Typically, you want to have some assert fail if the code does not work as expected. You can check some of the files in `tests` to see how that works.

pyroomacoustics/test_pra.py
```diff
@@ -0,0 +1,591 @@
...
+        sh = f.shape[0] - 1
+
+        k = (-1 / 360) * ((Sxx[sh][_] - Sxx[sh - 1][_]) / (f[sh] - f[sh - 1]))
+        grp_val.append(k)
+        grp_delay.append(grp_val)
+
+    g_delay = np.array(grp_delay).reshape(f.shape[0], -1)
+    return g_delay
+
+#/home/psrivast/PycharmProjects/axis_2_phd/Debug_sofa_file.sofa
```
fakufaku, line 44: remove

pyroomacoustics/test_pra.py
```diff
@@ -0,0 +1,591 @@
...
+#/home/psrivast/Téléchargements/AKG_c480_c414_CUBE.sofa
+#/home/psrivast/Téléchargements/LSPs_HATS_GuitarCabinets_Akustikmessplatz.sofa
+
+
+dir_obj_mic = DIRPATRir(
+    orientation=DirectionVector(azimuth=0,colatitude=0 , degrees=True),
+    path="/home/psrivast/Téléchargements/AKG_c480_c414_CUBE.sofa",
+    directivity_pattern=1,
+    fs=16000,
+    #no_points_on_fibo_sphere=0
```
fakufaku, line 55: remove

pyroomacoustics/test_pra.py
```diff
@@ -0,0 +1,591 @@
...
+
+
+
+
+dir_obj_sr = DIRPATRir(
+    orientation=DirectionVector(azimuth=0,colatitude=0, degrees=True),
+    path="/home/psrivast/Téléchargements/LSPs_HATS_GuitarCabinets_Akustikmessplatz.sofa",
+    directivity_pattern=2,
+    fs=16000,
+    #no_points_on_fibo_sphere=0
```
fakufaku, line 67: remove

pyroomacoustics/test_pra.py
```diff
@@ -0,0 +1,591 @@
...
+         [o[1] + w, o[1] + w, o[1] + w, o[1] + w, o[1] + w]]    # y coordinate of points in inside surface
+    z = [[o[2], o[2], o[2], o[2], o[2]],                        # z coordinate of points in bottom surface
+         [o[2] + h, o[2] + h, o[2] + h, o[2] + h, o[2] + h],    # z coordinate of points in upper surface
+         [o[2], o[2], o[2] + h, o[2] + h, o[2]],                # z coordinate of points in outside surface
+         [o[2], o[2], o[2] + h, o[2] + h, o[2]]]                # z coordinate of points in inside surface
+    return x,y,z
+
+
+
+#az=90,col=10
```
fakufaku, line 99: remove

fakufaku (changes requested):
@prerak23, thanks for the fantastic contribution! It will great to be able to load measured responses!

The contribution is quite large, so I expect it will take a little bit of time to integrate the whole thing into the code base. Please bear with me 🙏 

I have done a first pass on your code and I think I got a bunch of stuff. You can just correct the stuff that is obvious, commit and push to this branch to update the PR. For stuff that is not obvious we can continue the discussion in the PR.

Obvious stuff
- remove commented out lines
- remove imports of local paths
- convert docstrings to [numpydoc style](https://numpydoc.readthedocs.io/en/latest/format.html)
- format code with Black (a recent version)

Stuff that is not obvious:
- design, separation of tasks in `open_sofa_interpolate` file
- application of rotation of the source/receiver
- integration in room class

Also, I saw that your branch is not in sync with the latest version in master, so please start by merging with the latest version.

pyroomacoustics/room.py
```diff
@@ -2062,18 +2002,34 @@ def compute_rir(self):
...
 
+                """
+                Source Directivity : Obj of CardioidFamily , Receiver Directivity : Obj of freq_dependent_response_sofa 
+                Source Directivity : Obj of freq_dependent_response_sofa , Receiver Directivity : Obj of CardioidFamily
+
+                Check if the directivity object of source and receiver are of same class, if not throw error
+
+                """
+
+                if self.mic_array.directivity is not None and self.sources[s].directivity is not None:
```
fakufaku, line 2019: I just noticed the current code restricts the source/microphone to be of the same class (CardioidFamily *or* DIRPATir).
Ideally, we would like everything to be compatible. The way to do it is to create an interface common to all types of source/mic and hide the details behind.
Let's discuss how we can do this.

fakufaku:
@prerak23 I think we are close but
1) You have included 1600+ files (from `libroom/ext/eigen`) in the commit. Could you please remove them ? Also, before committing, I would suggest that you review what files you are about to commit with the `git status` command.
2) You need to add `import os` in the import section of your test files.

pyroomacoustics/room.py
```diff
@@ -98,38 +98,29 @@
...
     actually **not** the energy absorption we use now.  The ``absorption``
     parameter is now deprecated and will be removed in the future.
-    
-    
+
 Randomized Image Method
 ~~~~~~~~~~~~~~~~~~~~~~~~~
-
-In highly symmetric shoebox rooms, the regularity of image sources’ positions 
+In highly symmetric shoebox rooms, the regularity of image sources’ positions
```
fakufaku, line 104: Some blank lines are missing. Can we add them back ?

pyroomacoustics/room.py
```diff
@@ -617,12 +606,27 @@ def callback_mix(premix, snr=0, sir=0, ref_mic=0, n_src=None, n_tgt=None):
...
 from .utilities import angle_function
+from .directivities import CardioidFamily, source_angle_shoebox, Directivity, DIRPATRir
+
+from math import floor
+from timeit import default_timer as timer
+from scipy.fft import fft, fftfreq, ifft
+from scipy.signal import fftconvolve
+import sys
+
+# sys.path.insert(0, "/home/psrivast/PycharmProjects/axis_2_phd")
```
fakufaku, line 622: please remove such lines

pyroomacoustics/room.py
```diff
@@ -2165,137 +2202,450 @@ def compute_rir(self):
...
+                            air_decay = np.exp(-0.5 * air_abs * distance_rir)
+                            band[fdl2 : N + fdl2] *= air_decay
+
+                    # Sum up IR'S for all the bands
+                    np.sum(rir_bands, axis=0, out=ir)
+
+                else:
+
+                    """
+                    Checks if both source and microphone directivity is from the class DIRPATRir
```
fakufaku, line 2403: I think this comment is not relevant anymore.

pyroomacoustics/room.py
```diff
@@ -2165,137 +2202,450 @@ def compute_rir(self):
...
+                if (
+                    (
+                        self.mic_array.directivity is None
+                        or isinstance(self.mic_array.directivity[m], CardioidFamily)
+                    )
+                    and (
+                        self.sources[s].directivity is None
+                        or isinstance(self.sources[s].directivity, CardioidFamily)
+                    )
+                ) or self.simulator_state["rt_needed"]:
```
fakufaku, line 2284: why is the last condition needed ?

fakufaku: It may be easier to order this if clause the other way around. Now, there are only two ways to proceed:
1) If either mic or source is from DIRPAT class, we use the new way
2) If not we use the other way
Thus
```python
if isinstance(self.mic_array.directivity[m], DIRPATPattern) or isinstance(self.sources[s].directivity, DIRPATPattern):
    # new way
else:
    # old way
```

pyroomacoustics/room.py
```diff
@@ -2165,137 +2202,450 @@ def compute_rir(self):
...
+                            # is shorter than the delay to to the filter
+                            # hence: time + fdl2
+
+                            time_adjust = (
+                                time + fdl2 / self.fs
+                            )  # This remains the same for all octave bands.
+
+                            fast_rir_builder(
+                                ir_loc, time_adjust, alpha, vis, self.fs, fdl
+                            )
```
fakufaku, line 2335: There are some missing blank lines that make the diff difficult to understand.

pyroomacoustics/room.py
```diff
@@ -2165,137 +2202,450 @@ def compute_rir(self):
...
+                    cpy_ir_len_3[i, :],
+                    final_fir_IS_len,
+                )
+
+                ir_diff[time_ip : (time_ip + final_fir_IS_len)] += np.real(out)
+
+            """
+            IS method without receiver and source response , right now it's being handled by pyroom acoustics octave band method.
+            The below defined is an DFT based method to create RIR , but this does not take into account ray tracing as it is restricted by pyroom.
+            else:
```
fakufaku, line 2637: this seems like old commented out code. Please remove it.

pyroomacoustics/tests/test_dirpat_class.py
```diff
@@ -0,0 +1,114 @@
...
+PATTERN_MIC_DIRPAT_ID = "AKG_c480"
+ORIENTATION_MIC = DirectionVector(azimuth=90, colatitude=90, degrees=True)
+directivity_MIC = DIRPATRir(
+    orientation=ORIENTATION_MIC,
+    path=path_DIRPAT_file,
+    DIRPAT_pattern_enum=PATTERN_MIC_DIRPAT_ID,
+    fs=16000,
+)
+
+"""
```
fakufaku, line 86: remove commented out code.

pyroomacoustics/tests/test_dirpat_class_cardioid.py
```diff
@@ -0,0 +1,106 @@
...
+# directivity_SRC.set_orientation(70, 34)
+
+
+room.compute_rir()
+
+rir_1_0 = room.rir[0][0]
+
+
+
+# np.save("/home/psrivast/PycharmProjects/axis_2_phd/pyroom_acous_push_1.npy", rir_1_0)
```
fakufaku, line 106: remove commented out code.

pyroomacoustics/tests/test_dirpat_class.py
```diff
@@ -0,0 +1,114 @@
+import numpy as np
```
fakufaku, line 1: The test is not testing anything. It should have at least one method inside that is prefixed with `test_`. Such methods are run automatically. In the method we check that values returned by the new functions/code match some expected values. The check is done with an assert that should fail if an unexpected value is returned.

fakufaku (changes requested):
There is still quite some things to work out.

Please start by cleaning all the simple things: comments, blank lines, dead code, etc.

Then we can move on to the more serious stuff (tests, etc).

pyroomacoustics/acoustics.py
```diff
@@ -160,26 +163,27 @@ class OctaveBandsFactory(object):
         Use third octave bands if True (default: False)
     """
 
-    def __init__(self, base_frequency=125.0, fs=16000, n_fft=512):
+    def __init__(self, base_frequency=125.0, fs=16000, n_fft=128):
```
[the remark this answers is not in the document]
fakufaku: It seems the filter length was changed to obtain shorter lengths of the impulse response.
This is a valid reason, but should really be controlled by a new parameter of the Room objects.

pyroomacoustics/room.py
```diff
@@ -912,6 +933,11 @@ def _var_init(
         # initialize the attribute for the impulse responses
         self.rir = None
 
+        # self.sh_deg = 12
```
fakufaku, line 936: still some comments.

pyroomacoustics/room.py
```diff
@@ -2165,137 +2218,437 @@ def compute_rir(self):
...
+                    time = (
+                        dist / self.c
+                    )  # Calculate time of arrival for each image source
+                    t_max = (
+                        time.max()
+                    )  # The image source which takes the most time to arrive to this particular microphone
+                    N = int(
+                        math.ceil(t_max * self.fs)
+                    )  # What will be the length of RIR according to t_max
+                    print("Minimum Time", time.min() * self.fs)
```
fakufaku, line 2240: Remove print statement.

pyroomacoustics/room.py
```diff
@@ -2165,137 +2218,437 @@ def compute_rir(self):
...
+                            if self.sources[s].directivity is not None and isinstance(
+                                self.sources[s].directivity, CardioidFamily
+                            ):
+                                alpha *= self.sources[s].directivity.get_response(
+                                    azimuth=azimuth_s,
+                                    colatitude=colatitude_s,
+                                    frequency=bw,
+                                    degrees=False,
+                                )
+                                print("Csrc")
```
fakufaku, line 2323: remove print

pyroomacoustics/room.py
```diff
@@ -2165,137 +2218,437 @@ def compute_rir(self):
...
+                            if self.mic_array.directivity is not None and isinstance(
+                                self.mic_array.directivity[m], CardioidFamily
+                            ):
+                                alpha *= self.mic_array.directivity[m].get_response(
+                                    azimuth=azimuth_m,
+                                    colatitude=colatitude_m,
+                                    frequency=bw,
+                                    degrees=False,
+                                )
+                                print("Cmic")
```
fakufaku, line 2312: remove print

pyroomacoustics/room.py
```diff
@@ -2165,137 +2218,437 @@ def compute_rir(self):
...
+                Constructed RIR for this particlar src mic pair .
+
+            The constructed RIR still lacks air absorption and distance absorption because in the old pyroom these calculation happens on the octave band level.
+
+
+        """
+
+        attenuations = attenuations / dist
+        alp = []
+        window_length = 81
```
fakufaku, line 2488: the window lengths should be obtained from the global parameters
```python
window_length = constants.get("frac_delay_length")
```

pyroomacoustics/room.py
```diff
@@ -2165,137 +2218,437 @@ def compute_rir(self):
...
+
+                assert (
+                    self.fs == self.mic_array.directivity[mic_pos].fs
+                ), "Directivity mic frequency of simulation should be same as frequency of interpolation"
+
+                resp_mic = self.mic_array.directivity[mic_pos].get_response(
+                    azimuth=azi_m, colatitude=col_m
+                )
+
+        # else:
```
fakufaku, line 2560: remove commented code

pyroomacoustics/room.py
```diff
@@ -2165,137 +2218,437 @@ def compute_rir(self):
...
+            att_in_dft_scale_ = self.octave_bands.octave_band_dft_interpolation(
+                att_in_octave_band,
+                self.air_absorption,
+                dist[i],
+                att_in_dft_scale_,
+                bws,
+                self.min_phase,
+            )
+
+            #time_ip = int(floor(sample_frac[i]))  # Calculating the integer sample
```
fakufaku, line 2605: remove commented code.

fakufaku (changes requested):
I found some more comments and print statements to remove.

I think you changed something in the `if` statement to choose the method for the simulation (old or your new way) and it does not look right anymore.

PR #271, opened 2022-07-05
Using efficient np.linalg.multi_dot

maldil:
It is more succinct and effective to refactor code to use np.linalg.multi_dot rather than numerous np.dot routines.
What do you think about this change that has practical value?

- [ ] Are there docstrings ? Do they follow the [numpydoc] (https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) style ? **There are no docstrings**
- [ ] Have you checked that the doc builds properly and that any new file has been added to the repo ? How to do that is covered in the [documentation](https://pyroomacoustics.readthedocs.io/en/pypi-release/contributing.html#documentation). **Yes**
- [ ] Is there a unit test for the proposed code modification ? If the PR addresses an issue, the test should make sure the issue is fixed. **No, this is a refactoring.**
- [ ] Last but not least, did you document the proposed change in the CHANGELOG file ? It should go under "Unreleased". **No, this is a refactoring.**

Happy PR :smiley:

### DIFF
--- a/pyroomacoustics/beamforming.py
+++ b/pyroomacoustics/beamforming.py
@@ -1432,7 +1432,7 @@ class Beamformer(MicrophoneArray):
 
         # compute and return SNR
         num = np.inner(g_val.T, np.dot(R_xx, g_val))
-        denom = np.inner(np.dot(g_val.T, K_nq), g_val)                 
+        denom = np.inner(np.dot(g_val.T, K_nq), g_val)
 
         return num / denom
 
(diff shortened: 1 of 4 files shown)


fakufaku:
Hi @maldil , thank you so much for taking the time to open a pull request.

I didn't know about multi_dot. Thank you for pointing this out. It is new from numpy v1.19.0.

This would break compatibility for older versions of numpy, so I don't think we should use it, unless there is a large speed improvement for example.
Have you measured the runtime difference ?

fakufaku:
Right, I understand it should be faster because it will pick the optimal ordering of parenthesis. 
I'm more interested in the practical impact on the performance of DOA computations. If we can get 4.5x improvement overall on the DOA module that would be worth thinking about it.
Do you have a performance comparison on the DOA module ?

fakufaku:
I think you could quickly make one by timing the doa algorithms in the `examples/doa_algorithms.py` script. For runtime measurements, it is recommended to use `time.perf_counter`, repeat the computation you want to measure several times and take the mean or median.

PR #286, opened 2022-11-11
define a3 in room.py

usernamenoahfoster:
Fixed error, undefined object "a3"


fakufaku:
Thanks for reporting this. I will merge it asap.

PR #288, opened 2022-11-12
fix number mixed up

hutauf:
room was not L shaped


fakufaku:
Thanks for catching this!

PR #302, opened 2023-02-02
Support for recorded source/mic directivities from DIRPAT database (#259)

fakufaku:
This PR is to integrate the work done by @prerak23 into the master branch:

### Generalize for other SOFA files

The goal is to generalize what has been done for DIRPAT by @prerak23 to all SOFA files.
The target syntax is as follows.
```python
# the file reader object reads the file and optionally performs interpolation
# if the file contains multiple directivities, they are all read
hrtf = MeasuredDirectivityFile(
    path="hrtf_file.sofa",
    fs=fs,
    interp_order=12,
    interp_n_points=1000,
)

# orientations can be provided as rotation matrices
orientation = Rotation3D([colatitude_deg, azimuth_deg], "yz", degrees=True)

# we can then choose which directivities we want from the file
dir_left = hrtf.get_mic_directivity("left", orientation=orientation)
dir_right = hrtf.get_mic_directivity("right", orientation=orientation)
```

### TODO
- [ ] documentation
- [ ] changelog
- [X] Read multiple objects from a single file to be able to simulate mic arrays more smoothly, e.g. eigenmike. This could be implemented by detecting the multiple channels after the call to `get_response` that would return a `(n_image_sources, n_channels, n_taps)` array.
- [X] Implement 3D rotations with Euler angles

Co-authored-by: prerak23 <prerak23@github.com>

Thanks for sending a pull request (PR), we really appreciate that! Before hitting the submit button, we'd be really glad if you could make sure all the following points have been cleared.

Please also refer to the doc on [contributing](https://pyroomacoustics.readthedocs.io/en/pypi-release/contributing.html) for more details. Even if you can't check all the boxes below, do not hesitate to go ahead with the PR and ask for help there.

- [ ] Are there docstrings ? Do they follow the [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) style ?
- [ ] Have you run the tests by doing `nosetests` or `py.test` at the root of the repo ?
- [ ] Have you checked that the doc builds properly and that any new file has been added to the repo ? How to do that is covered in the [documentation](https://pyroomacoustics.readthedocs.io/en/pypi-release/contributing.html#documentation).
- [ ] Is there a unit test for the proposed code modification ? If the PR addresses an issue, the test should make sure the issue is fixed.
- [ ] Last but not least, did you document the proposed change in the CHANGELOG file ? It should go under "Unreleased".

Happy PR :smiley:

PR #317, opened 2023-06-05
Speech directivity

pkendrick:
This PR is an attempt to add frequency dependent source directivity. I have created a new class `SpeechDirectivity` and adds some examples and unit tests. This is my first PR for pyroomacoustics, let me know if ive missed anything or any issues. Thanks!

---

Happy PR :smiley:


fakufaku:
Hi @pkendrick , thank you so much for the PR!

I think it would be pretty nice to support such sources.

There are however a couple of roadblocks ahead
1. I am currently integrating another pull request to add support for measured directivites #302 , I think there is significant code overlap with your own contribution. In particular, I have significantly refactored the `compute_rir` function. The bad news is that you will probably have to rewrite that part so that it plays nicely with your contribution. The good news is that the refactored version should be a lot simpler to understand.
2. As part of the previous point, I have changed a little bit the concept of the `Directivity` object so that when requesting the gains at different angles, it will return the gains for all frequency bands, rather than having a `frequency` argument. The assumptions is that either the object is frequency independent, uses octave bands (same as simulation), or will return an impulse response.

I am quite advanced in integrating #302 so I don't think it should take too long before it is merged in master. Essentially, it is working and the tests are passing, but I would like to check a little bit more and possibly add extra tests and doc.

pkendrick:
Thanks alot for letting me know, that all sounds pretty good. No problem, I am happy to update what to match this work when it is ready. Cheers!

PR #335, opened 2024-01-15
moving sound source

kehinde-elelu:

Happy PR :smiley:

### DIFF
--- a/pyroomacoustics/room.py
+++ b/pyroomacoustics/room.py
@@ -2080,6 +2080,115 @@ class Room(object):
 
         return self.add(mic_array)
 
+    def simulate_moving_sound(
+        self,
+        position,
+        signal=None,
+        delay=0,
+        fs=None,
+        stept=200,
+        speed=5,
+        x_direction=True,
+        y_direction=True,
+    ):
+        """
+        Adds a moving sound source given by its position in the room.
+        Simulate all the locations the sound source is moving to
+        determining the RIR.
+        Perform time-varying convolution on the output signal.
+
+        Parameters
+        -----------
+        position: ndarray, shape: (2,) or (3,)
+            The starting location of the moving source in the room.
+        signal: ndarray, shape: (n_samples,), optional
+            The signal played by the source.
+        delay: float, optional
+            A time delay until the source signal starts
+            in the simulation.
+        fs: float, optional
+            The sampling frequency of the microphone, if different from that of the room.
+        stept: float, (The default is 200).
+            The step duration, measured in milliseconds.
+            It is used to determine the duration of each simulation
+            step in the virtual room.
+        speed: float, (The default is 5).
+            Speed of the moving sound source in the virtual room.
+            It is measured in meters per second.
+        x_direction, y_direction: bool, optional
+            Flags indicating the direction of movement. True for forward, False for backward.
+
+        Returns
+        -------
+        movemix: ndarray
+            audio signal obtained from the simulation.
+        filter_kernels: ndarray
+            Room impulse responses at each step.
+        """
+
+        # Validate parameters
+
+        if signal is None:
+            raise ValueError("Please provide a signal for the sound source.")
+
+        if fs is None or fs <= 0:
+            raise ValueError("Invalid sampling frequency (fs).")
+
+        # Calculate the number of samples in each step
+        stepn = int(stept * fs / 1000)
+        # Calculate the distance traveled in a step
+        stepd = speed * stept / 1000
+        # The number of simulation steps needed to process the entire audio signal.
+        n = int(len(signal) / stepn)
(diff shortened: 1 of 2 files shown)


kehinde-elelu:
@fakufaku I have created a pull request for the simulation of the movement of sound in a ShoeBox 
This is my first time contributing to GitHub and any changes, so in case I need to make some more changes, let me know. 

The last part still missing is creating a time-varying convolution. Quick question, should this be a separate function or added to the sound moving function?

pyroomacoustics/room.py
```diff
@@ -2080,6 +2080,115 @@ def add_microphone_array(self, mic_array, directivity=None):
...
+            self.sources = []
+            self.add_source(
+                position=source_location,
+                delay=delay,
+                signal=signal[i * stepn : (i + 1) * stepn],
+            )
+            # Simulate the room and obtain the room impulse response
+            self.simulate()
+
+            # only use the duration of original, to avoid echo
```
fakufaku, line 2176: For the time varying convolution, the tail (i.e., the samples beyond the original length) need to be kept and added to the head of the following segment.
Kind of like overlap add for long filtering (except the filter changes everytime).
You can think about it this way: after the source has moved, you still hear echoes generated by the source at the previous position.

pyroomacoustics/room.py
```diff
@@ -2080,6 +2080,115 @@ def add_microphone_array(self, mic_array, directivity=None):
...
+            # set data type as 16bit, ref:https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.write.html
```
fakufaku, line 2177: Please do not do this here.
The scaling, normalization, and data type casting should happen only when saving to file, which is independent from the simulation.

pyroomacoustics/room.py
```diff
@@ -2080,6 +2080,115 @@ def add_microphone_array(self, mic_array, directivity=None):
...
+            record_ = self.mic_array.signals[:, :stepn].astype("int16")
+
+            if i == 0:
+                movemix = record_
+                filter_kernels = self.rir.copy()
+            else:
+                movemix = np.concatenate((movemix, record_), axis=1)
```
fakufaku, line 2184: This is very inefficient as a copy of the array occurs at every loop, and at each loop the array becomes longer. In the end this incurs a cost quadratic in the length of the array.
The correct way to do it is to append to a list in the loop and concatenate all the bits at once at the end.

pyroomacoustics/room.py
```diff
@@ -2080,6 +2080,115 @@ def add_microphone_array(self, mic_array, directivity=None):
...
+                filter_kernels = np.concatenate(
```
fakufaku, line 2185: same

pyroomacoustics/tests/test_moving_sound.py
```diff
@@ -0,0 +1,42 @@
+import numpy as np
+
+import pyroomacoustics as pra
+
+
+def test_simulation_output_shape():
```
fakufaku, line 6: This is good as a basic test, but it would be nice to have some tests for the correctness of the time-varying convolution too.
One test for that is that the result of a time-varying convolution where all the filters are the same should be the same as a single long convolution.

pyroomacoustics/tests/test_moving_sound.py
```diff
@@ -0,0 +1,42 @@
...
+        position=position,
+        signal=signal,
+        fs=fs,
+        stept=stept,
+        speed=speed,
+        x_direction=x_direction,
+        y_direction=y_direction,
+    )
+
+    print(movemix, filter_kernels)
```
fakufaku, line 38: There should be some assert at the end of the test that fails is something is wrong with the code.

fakufaku (changes requested):
Thank you so much for starting this! The code is nice and well-commentd, bravo!

I think this is a good start, but there are a few things to change to make it the best possible.
I like that you keep things simple, but I can see two problems with this implementation.
1. It is merged as a method of the Room class, but this does not seem necessary.
2. The only movement allowed is a straight line.

I would like to suggest to change the code to modularize the different tasks involved.
1. Define trajectories (of sources, but also possibly microphones). We can have a collection of functions that all tackle different trajectories. The input would depend on the type of trajectory and the outputs would be
  a. a list of time stamps
  b. a list of locations
2. the simulation would be done by a function taking as input a `Room` object and the outputs of the above function for sources and microphones (both could be optional). Since the function would be outside of the `Room` class, it can be defined in a separate file, e.g. `pyroomacoustics/dynamic.py`.

Please do not hesitate to comment if some things are unclear.

kehinde-elelu:
Hello @fakufaku, notice another dev has created a pull request for a moving sound source. Does that mean it is no longer needed?

PR #401, opened 2025-05-18
added support for moving source

dzejnwashere:

Happy PR :smiley:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,14 @@ adheres to `Semantic Versioning <http://semver.org/spec/v2.0.0.html>`_.
 `Unreleased`_
 -------------
 
-Nothing yet
+During the testing PR moving sound source #335, I found some issues, which I tried to improve with this PR:
+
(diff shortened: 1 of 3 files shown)


fakufaku:
Hi @dzejnwashere , thanks for the contribution. I think there is a possiblity that the room impulse responses (RIR) are actually not recalculated between the steps.
This is not very well documented, but pyroomacoustics stores the room impulse responses, and does not recompute them after the simulation is done. So updating the source locations is not sufficient to guarantee the RIR are update.
I think this is why the signal scale was also not changed.
I apologize for the poor documentation on that part... 🙇 

What I would suggest is to precompute all the source locations in advance, and then simulate all the RIRs at once. This would solve the issue and probably also make the code simpler.

Another thing is that `room.py` is already a monster at the moment and if possible I would like to avoid adding the moving sources in there.
Also, I don't think it is necessary for the code to be a method of room to work.
At first, would you be ok to change this code to be an example and add in the `examples` folder?

fakufaku:
Finally, the code should be linted with `black` and `isort`.
Please consult [CONTRIBUTING](https://github.com/LCAV/pyroomacoustics/blob/master/CONTRIBUTING.rst#coding-style) for instructions.

PR #434, opened 2026-04-27
fix: clamp RT histogram to non-negative before sqrt

### LINKED ISSUE #433: NaN with raytracing scattering for mics close to the wall
I've been running into an occasional NaN issue when simulating with the new ray tracing scattering. I'm using Pyroomacoustics 0.10.0. Managed to boil it down to the following example.

The NaN occurs when negative values reach the np.sqrt in rt.py at:

```python
        # Impulse response for every octave band for each microphone.
        rir_bands[b, :] = seq_bp * np.sqrt(hist)
```

Only occurs when scattering materials are used with raytracing enabled and the receiver is closer to a wall than the receiver_radius.

Example where it occurs:
```python
import warnings

import numpy as np
import pyroomacoustics as pra

for receiver_radius in [0.5, 0.1]:  # the 0.1 case leaves the mic farther from the wall than the receiver_radius
    for use_scattering in [False, True]:
        pra.random.seed(140)  # this seed seems to reproduce the NaN issue reliably on my machine.

        room = pra.ShoeBox(
            [4.144901752471924, 6.039876461029053, 2.7347633838653564],
            fs=24000,
            materials=pra.Material(energy_absorption=0.13035287, scattering=0.1 if use_scattering else 0.0),
            max_order=0,
            ray_tracing=True,
            air_absorption=True,
        )
        room.set_ray_tracing(receiver_radius=receiver_radius)

        # The microphone is fairly close to the wall. Within the default receiver_radius (0.5 m)
        room.add_microphone([2.304156303866167, 0.17025121891320288, 1.4909059943866196])
        room.add_source([1.9948741012396916, 3.8261834557486685, 0.8465485818395717])

        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always")
            room.compute_rir()

            sqrt_warns = [x for x in w if "invalid value encountered in sqrt" in str(x.message)]
            has_nan = np.any(np.isnan(room.rir[0][0]))
            print(f"RT scattering: {use_scattering}. Receiver radius: {receiver_radius}. Sqrt warn: {len(sqrt_warns)}, RIR has NaN: {has_nan}")

```

Script output:
```
RT scattering: False. Receiver radius: 0.5. Sqrt warn: 0, RIR has NaN: False
RT scattering: True. Receiver radius: 0.5. Sqrt warn: 1, RIR has NaN: True
RT scattering: False. Receiver radius: 0.1. Sqrt warn: 0, RIR has NaN: False
RT scattering: True. Receiver radius: 0.1. Sqrt warn: 0, RIR has NaN: False

```

jguice:
## Summary

- `interp_hist()` can produce tiny negative values (~1e-31) from floating-point noise in linear interpolation of energy histograms
- `np.sqrt()` turns these into NaN
- Downstream IIR filters (`sosfiltfilt` in `compute_rir`'s high-pass) propagate a single NaN sample to the **entire** RIR, silently destroying it
- Fix: `np.sqrt(np.maximum(hist, 0.0))` — clamp before sqrt

Fixes #433.

## Reproduction

The issue triggers when ray tracing + HRTF directivity are combined. A specific source/mic geometry produces a histogram band where linear interpolation dips infinitesimally below zero. The resulting NaN cascades through the IIR high-pass filter to corrupt every sample in the RIR.

## Test plan

- [x] Verified fix eliminates NaN in reproduction case from #433
- [x] Verified fix eliminates NaN in our production workload (14-source binaural room render with KU100 HRTF)
- [x] L/R balance corrected from +8 dB bias to -0.07 dB after fix

### DIFF
--- a/pyroomacoustics/simulation/rt.py
+++ b/pyroomacoustics/simulation/rt.py
@@ -280,7 +280,11 @@ def compute_rt_rir(
         hist = interp_hist(hist_bands[b, :], N)
 
         # Impulse response for every octave band for each microphone.
-        rir_bands[b, :] = seq_bp * np.sqrt(hist)
+        # Clamp to non-negative: interp_hist can produce tiny negatives
+        # from floating-point noise in linear interpolation.  sqrt of a
+        # negative yields NaN which downstream IIR filters (e.g. the
+        # high-pass in compute_rir) then propagate to the entire RIR.
+        rir_bands[b, :] = seq_bp * np.sqrt(np.maximum(hist, 0.0))
 
     if air_abs_coeffs is not None:
         if rir_bands.shape[0] == 1:


fakufaku:
Thanks for the fix!

PR #435, opened 2026-04-30
Add boolean flag to check if walls should be plotted in a room

matija-marijan:
Hi! Thanks for the wonderful project.

This PR addresses a small annoyance when trying to plot an `AnechoicRoom` instance, where a [1, 1, 1] dummy room gets plotted, even though it is unnecessary.

I added a boolean flag `plot_walls` to `Room.plot()` which either creates or doesn't create the wall `PatchCollection`.

Corresponding unit tests which check for the existence of such `PatchCollection` in various scenarios are added under `test_room_plot.py`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,12 @@ adheres to `Semantic Versioning <http://semver.org/spec/v2.0.0.html>`_.
 `Unreleased`_
 -------------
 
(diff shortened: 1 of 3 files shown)


fakufaku (approved):
Thanks for the contribution! This should make it nicer to look at.

PR #437, opened 2026-05-15
spiral_2D_array now uses the center parameter

### LINKED ISSUE #413: Spiral 2D array not using the center parameter
The function spiral_2D_array is not using the parameter center. In the beamforming.py file you can find the function spiral_2D_array  and the center parameter is not use.

robrui:
The `spiral_2D_array` function accepted a `center` parameter but didn't use it — the array was always returned at the origin. All other array generators (`linear_2D_array`, `circular_2D_array`, `poisson_2D_array`) add the center offset correctly, so this brings the spiral function in line.

Fixes #413.

- [X] Have you checked that the doc builds properly and that any new file has been added to the repo ? How to do that is covered in the [documentation](https://pyroomacoustics.readthedocs.io/en/pypi-release/contributing.html#documentation).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,10 @@ adheres to `Semantic Versioning <http://semver.org/spec/v2.0.0.html>`_.
 `Unreleased`_
 -------------
 
-Nothing yet.
+Fixed
+~~~~~
+
+- `spiral_2D_array` now correctly uses the ``center`` parameter. Previously, it always returned the array at the origin. (`#413 <https://github.com/LCAV/pyroomacoustics/issues/413>`_).
 
 
 `0.10.1`_ - 2026-05-01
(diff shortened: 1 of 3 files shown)


fakufaku:
Thanks for contributing! 🤗

robrui:
Quick note on the CI failures — all 46 failing tests are in tests/directivities/ and trace back to a 403 from sofacoustics.org:

```
urllib.error.URLError: <urlopen error Failed to download http://sofacoustics.org/data/database/mit/mit_kemar_large_pinna.sofa>
```

The whole domain is returning 403, even the homepage. So unrelated to the spiral changes here. Might resolve on its own once the server is back up.

fakufaku:
Thanks for following up. Your test is passing so I'll merge it and fix the CI later. 👍

PR #440, opened 2026-07-06
Fix to_float32 producing nan/inf for unsigned-integer (8-bit WAV) input

gaoflow:
### Problem

`to_float32` turns unsigned-integer input into `nan`/`inf`:

```python
import numpy as np
from pyroomacoustics.utilities import to_float32

to_float32(np.array([0, 128, 255], dtype=np.uint8))
# RuntimeWarning: divide by zero
# array([nan, inf, inf], dtype=float32)
```

It scales integer input by `abs(np.iinfo(dtype).min)`. For **unsigned** integer
dtypes that minimum is `0`, so the division is by zero.

This is hit in practice by **8-bit WAV files**: `scipy.io.wavfile.read` returns
them as `uint8`, and they flow through `create_noisy_signal` → `to_float32`
(whose docstring notes the input is "typically obtained from a WAV file"),
silently turning the whole signal into `nan`/`inf`.

### Fix

Handle unsigned integers separately. Unsigned PCM is centered at `2**(bits-1)`,
so map it with `(data - midpoint) / midpoint` — matching the
libsndfile/soundfile `PCM_U8` convention and consistent with the existing
signed-integer scaling (`data / 2**(bits-1)`). Signed-integer and float inputs
are unchanged:

```python
>>> to_float32(np.array([0, 128, 255], dtype=np.uint8))
array([-1.        ,  0.        ,  0.9921875], dtype=float32)
```

### Tests

`tests/test_to_float32.py`: `uint8` full-range finiteness + exact `(x-128)/128`
values + midpoint→0.0; `uint16` range; an `int16` regression guard (output
unchanged); float passthrough; and an end-to-end `create_noisy_signal` on a
generated 8-bit WAV asserting finite output.

RED→GREEN: 3 of the 5 new tests fail on the current code (uint8, uint16,
end-to-end produce nan/inf); all 5 pass with the fix. Relevant suites
(`test_to_float32`, `test_metrics`, `test_hpf`, `test_resample`,
`test_bandpass_filterbank`) → 16 passed, no regressions. `ruff check` clean on
the changed lines.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Fixed
 ~~~~~
 
 - `spiral_2D_array` now correctly uses the ``center`` parameter. Previously, it always returned the array at the origin. (`#413 <https://github.com/LCAV/pyroomacoustics/issues/413>`_).
+- ``to_float32`` no longer produces ``nan``/``inf`` for unsigned-integer input (e.g. 8-bit WAV read as ``uint8``). Unsigned PCM is now centered and scaled correctly instead of dividing by zero.
 
 
 `0.10.1`_ - 2026-05-01
--- a/pyroomacoustics/utilities.py
+++ b/pyroomacoustics/utilities.py
@@ -181,7 +181,13 @@ def to_float32(data):
         `data` as float32.
     """
 
-    if np.issubdtype(data.dtype, np.integer):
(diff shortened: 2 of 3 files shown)


fakufaku (approved):
This is great. Thank you for the fix!

PR #443, opened 2026-07-16
Fix normalization of integer and silent signals

gaoflow:
## Summary

- promote integer-valued signals before peak normalization so NumPy can represent fractional results
- preserve zero-amplitude signals instead of dividing by zero and producing `nan`
- cover dtype limits, bit scaling, and normalized silent WAV output with regression tests

## Problem

`normalize()` currently performs division in place on a copy of the input. Integer arrays therefore raise `UFuncOutputCastingError`, while a silent floating-point signal is divided by a zero peak and becomes `nan`. The latter can propagate into files written by `MicrophoneArray.to_wav(norm=True)`.

## Testing

- `571 passed` with the full test suite
- `black --check pyroomacoustics/utilities.py tests/test_normalize.py`
- `isort --check-only --diff --profile black pyroomacoustics/utilities.py tests/test_normalize.py`


binggao1230:
The lint check failed before installing or running any linter. `actions/setup-python@v4` did not find Python 3.9 in the runner cache, attempted to download it, and received GitHub's Unicorn HTML error page. Local checks pass with the workflow versions (`black 26.1.0` and `isort 7.0.0`), and the full suite passes (571 tests). Fork authors cannot rerun an upstream workflow job, so I am documenting the infrastructure failure here: https://github.com/LCAV/pyroomacoustics/actions/runs/29541362105/job/87764109869

fakufaku (approved):
Thanks for the improvement!

fakufaku:
Python 3.9 has been deprecated 8 months ago, so it might be time to drop it. Thanks for documenting!

